Issue #144, opened 2015-02-24
Keyring access control in OS X


jaraco:
When a new password is created through the OS X backend it's done through the `security` command, and the new keyring item is therefore set to always allow access from the `security` command. It seems reasonable to me that the creating process should have access, but because keyring delegates to the security command there's a particular problem: `security` will happily dump the password you're saving for anyone else! This access can be manually revoked, at the cost of manually confirming access each run, but there's no way to connect access to your particular application. (I'm not sure how it's verified anyways, so maybe this is a moot point without a signed binary or something of that sort.)

---
- Bitbucket: https://bitbucket.org/kang/python-keyring-lib/issue/144
- Originally reported by: Sam Birch
- Originally created at: 2014-04-29T03:18:19.853

jaraco:
Previously (before rev 438) there was native implementation for mac os x backend with ctypes.
I prefer native call to KeychainAPI instead of subprocess because ownership belongs to calling process not to 'security' app.
But I'm not sure it will work in virtual env or other sandboxes.

---

Original comment by: Denis Averin

jaraco:
I also would prefer a ctypes implementation to a subprocess.

---

Original comment by: Jason R. Coombs

jaraco:
As a temporary workaround, you can pass an option to security add-generic-password to make it so it does not give the 'security' process "trusted" permissions

from the security manpage:

> By default, the application which creates an item is trusted to
>             access its data without warning.  You can remove this default
>             access by explicitly specifying an empty app pathname: -T "". If
>             no keychain is specified, the password is added to the default
>             keychain.

---

Original comment by: Mark Grandi

hryamzik:
@jaraco so is it possible to set up keyring to use the `-T ""` option?

P.S. is there any way to change the app name (security to a custom one)?

jaraco:
I'm not sure. I think I answered that question elsewhere - maybe in a pull request or different ticket. But I believe passing `-T ""` would cause the application to force an interactive prompt on every access, which violates a primary goal of keyring (making interaction unnecessary for credential access).

hryamzik:
@jaraco in fact it will just ask next time and if user click on "always allow" – it would go away forever. The issue here is that any app can call security framework and allowing it to access all the keys without prompt by default doesn't seem to be a good idea.

jaraco:
I'll experiment with this idea, get a feel for the user experience.

jaraco:
[...]

PR #163, opened 2015-05-19
OS X adjustments

chrisidefix:
Issue #144 details how any process can access the created passcode (through security) under Mac OS X. By explicitly NOT setting any allowed application to access the new item, any access now requires the user to input her passcode.

PS: I also added a .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,16 @@
+__pycache__/
+*.py[cod]
+
+.py[co]
+build/
+bin/
+parts/
+tags
+keyring.egg-info
+develop-eggs
+.installed.cfg
+*.eggs/
+*.egg-info/
+*.egg
+dist/
+.idea
--- a/keyring/backends/OS_X.py
+++ b/keyring/backends/OS_X.py
@@ -50,6 +50,7 @@ class Keyring(KeyringBackend):
                 '-a', username,
                 '-s', service,
                 '-w', password,
+                '-T', '', #do NOT authorize any application by default (especially not security itself!)
                 '-U',
             ]
             call = subprocess.Popen(cmd, stderr=subprocess.PIPE,


jaraco:
I would prefer not to require any interactive input. Although there's value in using a single password to manage many, a primary purpose of keyring is to provide programmatic access to a secure credential set, such that an application (GUI or command-line or cron script) can access the OS-protected keyring.

I would like for the user's already authenticated session to be the guard against unauthorized use of the keyring. I wouldn't mind this behavior being the default for a subprocess-based keyring if there were already implemented a direct API (such as through ctypes), but the direct API keyring would still be the default for OS X.

chrisidefix:
I see your point, but it makes using keyring very vulnerable to attacks in my opinion and opening these keyring items to be accessed by any script without requiring authorisation doesn't seem secure.

jaraco:
They aren't accessible by any script. They're only accessed by a script run in the user's authenticated context, which is already authorized by the user's login.

You're somewhat right that it's less secure than requiring a password on each entry, which is less secure than not using keyring at all and requiring each password to be entered uniquely.

If security is paramount, then using keyring may not be the best option - or using a custom keyring provider that requires a password entry on each invocation.

However, there are many use cases where the current implementation provides a suitable balance of convenience and security. Changing that expectation without consideration for those cases would trouble for those users.

The current implementation is also well-aligned with the Windows keyring implementation, which similarly relies on the user's authenticated context to grant access to the keyring. In Windows, multiple applications can access the keyring, such that a password for a given resource can be shared by different applications in different contexts. I use this feature in Windows to access passwords from Powershell, without Python or keyring being involved at all.

So it would be inappropriate to make this change via fiat. A change that drastically impacts the API like this should be introduced incrementally, as a separate keyring backend, and as that backend gains acceptance, it might be considered to be the default backend, especially if these referenced vulnerabilities were shown to be impactful.

Issue #145, opened 2015-02-24
Password exposure through process listing on OS X


jaraco:
Keyring delegates password setting to `security` on OS X. When a password is set the process is Popen'd with the password in the command. I think this exposes the password to `ps`, e.g. you can see the arguments to `sleep` here:

```
>>> import subprocess
>>> subprocess.Popen(['sleep', '100'])
```

```
$ ps
PID   TTY           TIME CMD
50559 ttys002    0:00.01 sleep 100
```

---
- Bitbucket: https://bitbucket.org/kang/python-keyring-lib/issue/145
- Originally reported by: Sam Birch
- Originally created at: 2014-04-29T03:27:46.155

jaraco:
Oh. That seems not so good security at all. Perhaps keyring can pass the password using the IO stream. Otherwise, it may be necessary to use the C API, which I believe had other bugs.

---

Original comment by: Jason R. Coombs

jaraco:
Another way might be to pass it through an environment variable, but I'm not sure that `security` will read from one, and variable expansion in the command doesn't solve the problem (as `ps` shows it after expansion).

---

Original comment by: Sam Birch

jaraco:
It looks like you could do it via stdin through `security`'s interactive mode:

```
$ security -i
security> add-generic-password -a 'TEST' -s 'TEST SERVICE' -w 'PASSWORD'
security> <EOF> 
```

You could even keep a process w/ interactive mode open to reduce the process startup overhead.

---

Original comment by: Sam Birch

jaraco:
FYI, #13 is where subprocess-based invocation was added to workaround an apparent security issue with virtualenvs. It was released with Keyring 0.8.

---

Original comment by: Jason R. Coombs

jaraco:
I'm inclined to back out the changes from #13 and re-open that ticket. Performing passwords in-process (using ctypes) seems like the proper solution. If Mac disallows access to those APIs for virtualenv-hosted Pythons, then Apple should fix that issue or provide a means for the user to do so. That may be asking too much from Apple.

---

Original comment by: Jason R. Coombs

jaraco:
Thanks for the context Jason.

FWIW I'd rather have it working in virtualenv than proper, unfortunate as it is. I don't know all the details of the restrictions w/r/t the keychain in virtualenvs, but it sounds to me like the present policy is the "proper" one from Apple's perspective and they're unlikely to change it if so.

[...]

PR #165, opened 2015-06-19
Use security interactively instead of directly.

rbranche:
Use the security interactive prompt for OSX instead of calling
add-{}-password directly so the password is not in ps output.

### DIFF
--- a/keyring/backends/OS_X.py
+++ b/keyring/backends/OS_X.py
@@ -43,18 +43,18 @@ class Keyring(KeyringBackend):
             username = ''
         set_error = PasswordSetError("Can't store password in keychain")
         try:
-            # set up the call for security.
+            # Use the interactive security prompt, so the password is not in
+            # the ps output
             cmd = [
                 'security',
-                SecurityCommand('add', self.store),
-                '-a', username,
-                '-s', service,
-                '-w', password,
-                '-U',
+                '-i'
             ]
-            call = subprocess.Popen(cmd, stderr=subprocess.PIPE,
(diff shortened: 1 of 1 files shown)


jaraco:
@joneskoo I see you mentioned in #145 that -i was only added recently in 10.10. Yikes, that's recent.

Also, the tests fail with this PR (with non-ascii characters, I think), so although I've accepted it, I've pushed it into the `security-interactive` branch (for reference).

In the master branch, I'm adding OS X to the Travis CI tests, which should make it easier to see test acceptability.

joneskoo:
@jaraco Sorry, I'd love to tell you for a fact which version is the first to work with stdin password input but I don't have the facts at hand. https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/security.1.html man security(1) says it's for 10.9 and it has interactive -i so maybe I'm wrong.

I don't have notes on what would go wrong with 10.9 and I have no 10.9 available. http://docs.travis-ci.com/user/osx-ci-environment/ says they run 10.9.5.

jaraco:
I had to e-mail Travis to request the OS X feature be enabled. If keyring works with 10.9, that may be an old enough version to drop support for older OS X versions.

jaraco:
I tried testing everything under Travis with OS X, but it's still an experimental feature and a bit qwirky, so I've given up on the Travis on OS X bit. I think what I'll do since I run OS X now is to just run keyring with the PR code and assure myself it works suitably.

Issue #170, opened 2015-08-12
Python2 vs Python3 and PyQt4 vs PyQt5


jim-easterbrook:
I started porting an application that uses keyring from PyQt4 to PyQt5, but found it wouldn't run under Python2 because the kwallet backend uses PyQt4: `RuntimeError: the PyQt4.QtCore and PyQt5.QtCore modules both wrap the QObject class`.
I was surprised to find it does work under Python3 though. A bit of further investigation found that under Python3 I'm getting a different keying backend:

``` python
jim@brains:~$ python2
Python 2.7.8 (default, Sep 30 2014, 15:34:38) [GCC] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import keyring
>>> keyring.get_keyring()
<keyring.backends.kwallet.Keyring object at 0x7f61bdc8d850>
>>> 
jim@brains:~$ python3
Python 3.4.1 (default, May 23 2014, 17:48:28) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import keyring
>>> keyring.get_keyring()
<keyring.backends.Gnome.Keyring object at 0x7fe5246da320>
>>> 
jim@brains:~$ 
```

I'm running OpenSUSE 13.2 (KDE) but I'd like my application to be cross platform. I suspect kwallet's use of Qt4 is beyond our control, so should keyring choose a different backend if PyQt5 has already been imported?

jim-easterbrook:
I've done a bit more reading and found that KDE dialogs use Qt4, and can't be used from a Qt5 application. KDE is moving to Qt5, so in future we'll have the reverse problem.

jim-easterbrook:
If I install the KDE framework integration plugins I get the same behaviour in Python3 as in Python2, eliminating one of my variables. Now I just need to stop python keyring using the kwallet backend when using Qt5.
http://api.kde.org/frameworks-api/frameworks5-apidocs/frameworkintegration/html/index.html

mitya57:
The Kwallet backend should be definitely ported to PyQt5/PyKDE5. Can you do a pull request for this? :)

jim-easterbrook:
I've not found a Python KDE5 interface in the OpenSUSE repos, so using the KDE5 version of kwallet may not yet be possible. I'll have a go at making the kwallet backend fail more gracefully under Qt5 though.

mitya57:
[...]

PR #171, opened 2015-08-12
Kwallet backend fails gracefully if PyQt5 in use

jim-easterbrook:
This assumes PyQt5 is imported before keyring. Using import order like
this is not good, but how else can a module know if it's in a PyQt4 or
PyQt5 application?
See issue #170 for discussion prompting this change.

### DIFF
--- a/keyring/backends/kwallet.py
+++ b/keyring/backends/kwallet.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 import os
+import sys
 
 from ..py27compat import unicode_str
 from ..backend import KeyringBackend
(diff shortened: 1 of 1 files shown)


jaraco:
Thanks for this. 

Why is it that the PyQT5 section imports QtGUI but not kwallet?

jim-easterbrook:
I assumed QtGui needs to be defined as it's referred to elsewhere in the module. (I'm not clear how failure of a backend to load is detected by whatever chooses a backend.) I also left it as a hook for a future Qt5 compatible kwallet backend.

PR #172, opened 2015-09-06
Add plain text AWS S3 backend

germangh:
Very similar to the File backend but using Amazon's S3 as storage. A use case of this type of backend is to share credentials with AWS resources that have been associated to an IAM role that has access to the keyring's S3 bucket.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+__pycache__
+.eggs
+keyring.egg-info
--- a/README.rst
+++ b/README.rst
@@ -103,6 +103,18 @@ directory in the project checkout::
     keyring-path=/home/kang/pyworkspace/python-keyring-lib/demo/
 
 
+S3 backend configuration
+------------------------
+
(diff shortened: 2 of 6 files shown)


jaraco:
Thanks for the thorough implementation.

I have been lately very resistant to adding new backends to the core library, primarily because I hope to eventually move the core functionality (with the main OS-based backends) into the stdlib. I especially want to avoid adding particularly specialized backends into the library. However, the S3 backend does strike me as something that might have broader adoption.

Have you considered first creating a separate library, something like 'keyring-s3' or 'keyring-aws', which could (a) depend on keyring and (b) expose an entry point to make the keyring available? Then, anybody who wants this functionality could simply require 'keyring-s3' (or similar) and would have it. Then it could have its own namespace and maintenance flow.

germangh:
It makes sense to keep the S3 backend separate from the keyring library. So I'll do what you suggest and simply create a separate library and expose an entry point to the keyring library. 

I hope you manage to get keyring to become part of the stdlib. It's a really useful module.

PR #174, opened 2015-09-23
Add kwallet_dbus backend

jim-easterbrook:
A new backend that uses D-Bus to communicate with the KWallet daemon, avoiding dependency on PyQt4/5 and PyKDE. The existing kwallet backend remains for systems with Qt but without D-Bus.

### DIFF
--- a/keyring/backends/kwallet_dbus.py
+++ b/keyring/backends/kwallet_dbus.py
@@ -23,6 +23,12 @@ class Keyring(KeyringBackend):
     def priority(cls):
         if not dbus:
             raise RuntimeError('python-dbus not installed')
+        # make sure kwalletd is accessible
+        bus = dbus.SessionBus()
+        try:
+            remote_obj = bus.get_object('org.kde.kwalletd', '/modules/kwalletd')
+        except dbus.DBusException:
+            raise RuntimeError('cannot connect to org.kde.kwalletd')
         return 5.1
 
     def __init__(self, *arg, **kw):


jaraco:
This looks great.

Is there any reason this implementation shouldn't simply replace the kwallet implementation? It sure would be nice to have just one to maintain.

If not, could you at least add some tests such that this backend is tested?

jim-easterbrook:
I don't know how common it is to use KWallet without D-Bus (or even if it's possible), and didn't want to break any currently working systems.
I don't know anything about writing tests (or testing my tests!) so this might be easier for someone else to do.

jaraco:
I see that this implementation supplies an `appid`, whereas the other kwallet backend does not. Also, I don't think sys.argv can be reliably expected to reflect the application. Is it possible that the appid could be omitted to have consistent behavior with the other backend?

jim-easterbrook:
The `appid` is only provided to make the pop-up dialog asking the user to unlock the keyring a bit more meaningful. Using sys.argv provides a "best guess" of what program is wanting access, otherwise the dialog is anonymous. Is there a better way to determine the application name?

jaraco:
> Is there a better way to determine the application name?

There is not. The keyring interface does not solicit such a value. Perhaps it would be worth providing an interface for the application to provide an application name, but for now, I recommend leaving that blank or using a generic application name.

I'll see if I can throw together a test and address the concerns I have and get this integrated.

jaraco:
I've pushed the commits to the pr-174 branch in the repo, and made some adjustments. In order to merge this, I need the following:
1. Run the tests on a system with KDE and DBus, ensure the tests run and pass.
2. On a system with both Qt and DBus (such that tests pass on both keyrings), using the first keyring to save passwords, then use the other keyring to retrieve saved password, then do the same in the other direction (show at least basic compatibility across the keyrings).

Can you do those two things for me?

PR #178, opened 2015-12-07
fixes #177, gnome-keyring backend broken in fedora 23

### LINKED ISSUE #177: Can't use keyring in fedora 23
I have just installed fedora 23 and keyring package is broken here. If you try to use `set_password` method, it just hangs the application without any errors:

```
import keyring

keyring.set_password("myapp", "user", "password")
```

Python 2.7.10, keyring 5.6

dechandler:


jaraco:
Unfortunately, this change seems to have broken the tests. Somewhat to my dismay, these keyring backend modules must be importable even when there prerequisites aren't yet present. I should be able to adapt the patch to make it work.

PR #182, opened 2016-01-08
check for DISPLAY before start DBus Session

tuxlife:
Fix #174 for https://github.com/zalando-stups/zign/issues/9

### DIFF
--- a/keyring/backends/kwallet.py
+++ b/keyring/backends/kwallet.py
@@ -130,6 +130,8 @@ class DBusKeyring(KeyringBackend):
     def priority(cls):
         if 'dbus' not in globals():
             raise RuntimeError('python-dbus not installed')
+        if "DISPLAY" not in os.environ:
+            raise RuntimeError("cannot connect to X server")
         # make sure kwalletd is accessible
         bus = dbus.SessionBus()
         try:


jim-easterbrook:
As it's the `dbus.SessionBus()` call that raises an exception in your problem instance wouldn't it be easier just to move it into the `try: except:` block? (Just in case there are systems where D-bus exists but $DISPLAY doesn't.)

tuxlife:
try/except is possible, but a bad solution.

`dbus.SessionBus()` starts dbus, and this is without DISPLAY not possible.

jim-easterbrook:
You have a better knowledge of dbus than I do. I did not know it always requires $DISPLAY.

tuxlife:
@jim-easterbrook No, but i can read Error messages :smiley: 
@jaraco thanks!

jim-easterbrook:
I'd worry about generalising to all dbus installations from one particular system. I'm sure dbus can run on headless servers.

jaraco:
The Gnome backend seems to indicate that the presence of one of DISPLAY or DBUS_SESSION_BUS_ADDRESS is required. Perhaps that technique should be extracted and applied to any DBus implementation. I'll do that.

jaraco:
Hmm. Actually, I think I'm mistake. I don't think it's _or_, but rather _and_ -- both are required.

jaraco:
Thoughts on re-using the same detection machinery from Gnome.Keyring ^ ?

mitya57:
`$DISPLAY` is **not** required for D-Bus to work.

To confirm this, try running `dbus-run-session -- echo $DISPLAY` in a headless environment — it will print nothing, however D-Bus will work.

Checking for `$DBUS_SESSION_BUS_ADDRESS` is OK with traditional D-Bus, though it probably won't be exported with kdbus. So the best way, in my opinion, is a `try…catch` statement. See [here](https://github.com/mitya57/secretstorage/blob/master/secretstorage/__init__.py#L49) for how it's handled in my SecretStorage module.

jaraco:
Thanks @mitya57 and @jim-easterbrook for the comments. Your assessments led me to believe that the try/accept model would be better. I've released 7.0 with a fix to include that change (the major version bump due to another, unrelated change).

mitya57:
Thanks @jaraco!

PR #184, opened 2016-01-09
Add gitignore

mindw:
upstream is in an odd state (7.0 commit is not part of master).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,59 @@
+# Created by https://www.gitignore.io
+
+### Python ###
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+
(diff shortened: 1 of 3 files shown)


jaraco:
I dislike these spammy .gitignore files. If the settings are something that apply globally, then they should exist globally, and not copied and separately managed in every project. The .gitignore for a project should reflect the project's specific behavior. I've rebased this commit onto the master, so it's there, even though I don't like it.

mindw:
Sure. A minimal one will do?

mindw:
Sorry about that. 

I'll keep this mind for any future pulls for your projects.

jaraco:
No biggie. Just a pet peave.

PR #185, opened 2016-01-09
Stop relying on env variables for GNOME Keyring presence check

mitya57:
Call `get_default_keyring_sync()` and check the result instead.


jaraco:
Very nice. Thanks.

Issue #188, opened 2016-01-10
OS X backend not working when called from cron


n8henrie:
It was [recently brought to my attention](https://github.com/n8henrie/pycookiecheat/issues/8) that `python-keyring` doesn't appear to be able to retrieve a password from the OS X Keychain when called from cron. Any ideas why?

I assumed it has to do with the keychain being locked, but I'm able to access the keychain from cron using the `security` program (which keyring uses as its default OS X backend), as long as my cron entry first unlocks the keychain, e.g. `security unlock-keychain -p MYPASS /path/to/login.keychain && security find-generic-password -a ACCOUNT -s SERVICE -g >> outfile.log`

Also -- totally irrelevant, but -- what prompted the change to GitHub from Bitbucket? I didn't seen any discussion of it with the relevant commit, maybe I missed it.

jaraco:
I did respond to this issue with a detailed explanation and questions, but I now find (a) my response was lost and (b) my browser form memoizer doesn't have any record of my response. :(

Short answer - keyring can't guess `MYPASS`, so if unlocking the keychain is required by the `security` command, it will be required for Python keyring as well.

n8henrie:
Thanks a bunch for the quick response.

I'm not sure that I communicated the issue properly:
- If I **do not** unlock the keychain, I _cannot_ access passwords from cron using the `security` command, e.g. `* * * * * security find-generic-password -a foo -s bar -g >> out.log 2>&1`
- If I **do** unlock the keychain, I _can_ access passwords from cron using the `security` command, e.g. `security unlock-keychain -p MYPASS /path/to/login.keychain && security find-generic-password -s "foo" -a "bar" -g /path/to/login.keychain >> /path/to/fake.log 2>&1`
- **Even if** I unlock the keychain, I am unable to access the password with keyring, e.g. 
  - `security unlock-keychain -p MYPASS /path/to/login.keychain && /usr/local/bin/python3 -c 'import keyring; print(keyring.get_password("foo", "bar"))' >> /path/to/fake.log 2>&1`
  - `security unlock-keychain -p MYPASS /path/to/login.keychain && /usr/local/bin/keyring get "foo" "bar" >> /path/to/fake.log 2>&1`

Looking into it a bit further, I think it may actually be an issue with the default keychain, as demonstrated by examining what happens when we look at the `default-keychain` when called by `subprocess` (like keyring does) in a regular terminal vs in cron:

``` bash
[...]

PR #190, opened 2016-01-11
Add ability to specify keychain for OS X backend

n8henrie:
https://github.com/jaraco/keyring/issues/188

By default, the OS X backend uses the built-in `security` command to
store on retrieve passwords from the OS X keychain.

The `security` command, in turn, defaults to a user's
`~/Library/Keychains/login.keychain`. However, in certain settings such
as being called from `cron`, it will instead default to
`/Library/Keychains/System.keychain`, which will obviously not have the
user's passwords. It seems to use this as default even when being called
as a regular non-root user.

Many of the `security` subcommands have the flexibility to accept an
argument that specifies a non-default keychain. This commit gives the
default OS X Keyring backend `get_password` method the ability to accept
a keyword argument `keychain`. If given, this provides a specific
keychain to use and permits `python-keyring` to function in unusual
environments such as `cron`, without impairing its other functions.

### DIFF
--- a/keyring/backends/OS_X.py
+++ b/keyring/backends/OS_X.py
@@ -99,7 +99,7 @@ class Keyring(KeyringBackend):
         stdoutdata, stderrdata = call.communicate()
         return call.returncode
 
-    def get_password(self, service, username):
+    def get_password(self, service, username, keychain=None):
         if username is None:
             username = ''
         try:
@@ -111,6 +111,8 @@ class Keyring(KeyringBackend):
                 '-a', username,
                 '-s', service,
             ]
+            if keychain:
+                cmd.append(keychain)
             call = subprocess.Popen(
                 cmd,
                 stderr=subprocess.PIPE,
--- a/keyring/backends/OS_X.py
+++ b/keyring/backends/OS_X.py
@@ -24,6 +24,10 @@ class SecurityCommand(unicode_str):
 class Keyring(KeyringBackend):
     """Mac OS X Keychain"""
 
+    def __init__(self):
+        self.keychain = None
+
+
     # regex for extracting password from security call
     password_regex = re.compile("""password:\s*(?:0x(?P<hex>[0-9A-F]+)\s*)?"""
                                 """(?:"(?P<pw>.*)")?""")
@@ -111,6 +115,8 @@ class Keyring(KeyringBackend):
                 '-a', username,
                 '-s', service,
             ]
+            if self.keychain:
+                cmd.append(self.keychain)
             call = subprocess.Popen(
                 cmd,
                 stderr=subprocess.PIPE,
@@ -162,3 +168,17 @@ class Keyring(KeyringBackend):
                 raise del_error
         except:
             raise del_error
+
+    def set_keychain(self, keychain):
+        """Allows use of non-default keychain.
+
+        args:
+        keychain -- Absolute path to preferred keychain
+
+        OS X backend uses the ``security`` command, which can accept an
+        argument to specify a non-default keychain (default is usually
+        ``~/Library/Keychains/login.keychain``). This method allows keyring to
+        take advantage of this to also use non-default keychains if needed.
+        """
+
+        self.keychain = keychain
--- a/keyring/core.py
+++ b/keyring/core.py
@@ -33,10 +33,10 @@ def get_keyring():
     return _keyring_backend
 
 
-def get_password(service_name, username, **kwargs):
+def get_password(service_name, username):
     """Get password from the specified service.
     """
-    return _keyring_backend.get_password(service_name, username, **kwargs)
+    return _keyring_backend.get_password(service_name, username)
 
 
 def set_password(service_name, username, password):
--- a/keyring/core.py
+++ b/keyring/core.py
@@ -33,10 +33,10 @@ def get_keyring():
     return _keyring_backend
 
 
-def get_password(service_name, username):
+def get_password(service_name, username, **kwargs):
     """Get password from the specified service.
     """
-    return _keyring_backend.get_password(service_name, username)
+    return _keyring_backend.get_password(service_name, username, **kwargs)
 
 
 def set_password(service_name, username, password):


jaraco:
Thanks for sending the pull request.

One problem with this approach is that it's touching the interface for all keyrings. I feel a little uneasy adding support for arbitrary keyword arguments to (just) the get_password interface.

And this makes me wonder if there's an alternative approach that's more specific to this implementation or issue. I imagine, for example, the OS_X.Keyring class could have a 'set_keychain' method, that if called would affect which keychain is used. Then, in your code you could call `keyring.core.get_keyring().set_keychain('/Users/username/login.keychain')`.

Or even better, if this issue only affects cron, perhaps there's an environment variable or technique that could be invoked to address the behavior specifically in cron. In a brief search, I have found [this answer](http://stackoverflow.com/a/9844840/70170) and others from that link that suggest that `su -l` or (in Jenkins) a `SessionCreate` parameter causes the session to be opened as the logged-in user, giving the behavior that one finds in a login session. Perhaps the cron could be tricked into doing the same thing, saving both the code and usage from unnecessary complication.

mitya57:
How about specifying that name in Keyring's configuration file in a section specific to OS X backend?

Then we can allow other backends to store their settings there too, i.e. for the Secret Service backend it could be possible to specify the collection alias.

n8henrie:
Yeah, I agree that I didn't much like adding the `**kwargs` to the global `get_password` function (I suppose I could have limited it to just `keychain=None`).

I figured the best solution would be to replicate whatever environment `security` usually uses, so I searched for hours and tried every combination of `security default-keychain` and `security unlock-keychain` I could come up with. I'm not having any luck with `sudo -i` or `sudo -s`; `sudo -l` doesn't look correct:

>  -l, --list  If no command is specified, list the allowed (and forbidden) commands for the invoking user (or the user specified by the -U option) on the current host.  A longer list format is used if this option is
>                  specified multiple times and the security policy supports a verbose output format.

```
             If a command is specified and is permitted by the security policy, the fully-qualified path to the command is displayed along with any command line arguments.  If command is specified but not allowed,
             sudo will exit with a status value of 1.
```

A specific OS X Keyring may work if I don't find another way to just set up `security`. Though that said, I can also imagine other use cases in which a user may want to use `keyring` to access a non-default Keychain, so if a `set_keychain` method could be implemented without much hassle perhaps that might be more useful in the long run.

mitya57:
In my opinion, the keychain name is something that should be not part of API, but rather configurable by users. So an environment variable or a configuration file (like I proposed above) will work.

Re sudo: did you try the `-u` flag? I.e.:

```
root@lenovo:~# sudo -u dmitry whoami
dmitry
```

n8henrie:
Yes, I tried `sudo -u`. The other problem is that I'd prefer a solution that doesn't require root credentials, especially since there's no TTY available to input them for `sudo`. As I've already shown, cron can be run as a regular user, and if that user is logged in, `keyring` can access the keychain just fine if it is directed to the proper keychain -- no need for sudo, or `security unlock-keychain`, or any of that.

I'm not sure about a config file. It sounds like unnecessary hassle to me, since this is overall a rare use case, and I'm sure many OS X users like myself have no other reason for a keyring config file; the keyring defaults work great. It seems like a simple method to allow users to specify a keychain if they want to use a non-default one would resolve this issue, would hopefully not require changing any other existing code, and wouldn't require any additional setup or config files.

n8henrie:
Updated and removed the `**kwargs` stuff from `core.py`.

Added the `set_keychain` method to the OS X backend, and an instance variable `keychain`. Also added an `__init__` method to initialized it to `None` -- seemed cleaner than `try: except NameError` or adding a second `get_keychain` method, though I'm sure you all know better than me if those would be preferable.

All existing tests passing; I wasn't sure how to add a test for this method, but it doesn't seem to have broken anything as far as I can tell.

Also confirmed working with `cron` after `keyring.get_keyring().set_keychain("/path/to/login.keychain")`:

``` cron
* * * * * /usr/local/bin/python3 -c 'import keyring; print("Regular"); print(keyring.get_password("foo", "bar"))' >> /path/to/fake.log 2>&1
* * * * * /path/to/venv/bin/python3 -c 'import keyring; print("With set_keychain"); keyring.get_keyring().set_keychain("/path/to/login.keychain"); print(keyring.get_password("foo", "bar"))' >> /path/to/fake.log 2>&1
```

``` bash
$ cat fake.log
With set_keychain
MY_PASSWORD
Regular
None
```

jaraco:
Cool. I like it. Thanks.

jaraco:
I took some liberty to simplify the implementation by honoring the 'keychain' attribute and allowing it to be set. Can you test with 7.2 and confirm it works for your case?

n8henrie:
Sounds good, confirmed working with `keyring.get_keyring().keychain = 'foo'`. My only concern is that -- because this requires `.get_keyring()`, it may be easy for users to try to directly set `keyring.keychain = 'foo'`and not understand why things aren't working (i.e. there's no exception, appears to work). With the method (`keyring.set_keychain("foo")`), they would at least get `AttributeError: module 'keyring' has no attribute 'set_keychain'` to let them know something had gone awry.

Should be fine either way with a little documentation. Speaking of which, I went to add some on this but didn't think the frontpage README seemed like the right place for such a minor feature, so I just wrong long docstrings. I don't see the project at RTD and http://pythonhosted.org/keyring/ is bare and from 2009 -- are there docs somewhere that I'm missing? I had thought there were quite a bit of them prior to the GitHub migration.

jaraco:
> it may be easy for users to try to directly set keyring.keychain = 'foo'and not understand why things aren't working (i.e. there's no exception, appears to work)

I guess that issue isn't unique to this implementation, but is characteristic of any system that prefers simple attribute access to getter/setter methods (as seen in answers to [this question](http://stackoverflow.com/questions/165883/python-object-attributes-methodology-for-access)).

However, on further consideration, for a public API, it may be preferable to use explicit functions or methods for the reasons you raise.

Given the limited adoption I expect this feature to have, I'm inclined to leave it as-is, but I'll certainly consider it for future API aspects.

> I went to add some [documentation]

http://pythonhosted.org/keyring/ is current despite the date and updated with each release. It does need to be updated to pull in more of the docs from the modules. The source is in the repo as docs/. You're welcome to expand the docstring on the attribute. I acknowledge that I truncated the documentation from your original submission to focus on what it does and less on why someone would use it.

n8henrie:
All reasonable points, thanks for the thorough replies and consideration. I'm excited just to have had the opportunity to make a tiny contribution!

I saw `docs/` in the repo but assumed it was a work in progress as http://pythonhosted.org/keyring doesn't seem to have anything but the changelog. Is that right? For example, for me the [keyring module documentation](http://pythonhosted.org/keyring/index.html#module-keyring) just shows the word `__init__.py`, with no other submodules or docstrings or the readme or anything.

jaraco:
Yes. It's a work in progress - basically just a placeholder for now, but feel free to flesh it out.

PR #191, opened 2016-01-11
Format error message

ryuslash:
This morning I got the following error message which wasn't altogether very helpful:

```
RuntimeError: Keyring config exists only in the old location {config_file_old} and should be moved to {config_file_new} to work with this version of keyring.
```

With this change I get the proper error message, stating I should move it from `.local/share/python-keyring` to `.config/python-keyring`.

I tried this with both Python 2.7 and 3.5.


jaraco:
Yep. My mistake. Thanks for the patch.

jaraco:
Released as 7.1.2.

ryuslash:
Thanks!

PR #193, opened 2016-01-12
make sets backwards compatible with python 2.6

dandye:
This is the only change required to get `keyring` installed onto a centos6 box (which uses python 2.6)


jaraco:
keyring has dropped support for Python 2.6, in part to be able to accept the more elegant syntax and features afforded by Python 3.2/2.7 such as this one. I expect the library to start preferring importlib to `__import__` as well. Older distributions such as centos6 should use Keyring 5.7.1.

Issue #196, opened 2016-01-25
Error after upgrading to keyring 8: "No recommended password was available"


hjacobs:
Keyring 8 is broken for me (upgraded from PyPI today, keyring 7 worked before). I'm using Xubuntu 15.10 (Xfce).

We just call the usual `keyring.get_password('mai', ring_user)` ("mai" is our app).

Stacktrace:

```
$ mai
Traceback (most recent call last):
  File "/usr/local/bin/mai", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.4/dist-packages/mai/cli.py", line 305, in main
    cli()
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 1037, in invoke
    return Command.invoke(self, ctx)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/mai/cli.py", line 55, in cli
    login_with_profile(ctx.obj, profile, data.get(profile), awsprofile)
  File "/usr/local/lib/python3.4/dist-packages/mai/cli.py", line 231, in login_with_profile
    saml_xml, roles = saml_login(user, url)
  File "/usr/local/lib/python3.4/dist-packages/mai/cli.py", line 201, in saml_login
    saml_password = keyring.get_password('mai', ring_user)
  File "/usr/local/lib/python3.4/dist-packages/keyring/core.py", line 42, in get_password
    return _keyring_backend.get_password(service_name, username)
  File "/usr/local/lib/python3.4/dist-packages/keyring/backends/fail.py", line 18, in get_password
    raise RuntimeError("No recommended password was available")
RuntimeError: No recommended password was available
```

hjacobs:
Simple way to reproduce:

```
$ python3 -c "import keyring; keyring.get_password('test', 'jdoe')"
[...]

PR #200, opened 2016-01-27
README: Add a section on using Keyring on headless Linux systems

mitya57:
As the title says (and as requested in #196).


jaraco:
Awesome. This small patch makes keyring _so_ much better.

PR #202, opened 2016-01-28
fix syntax of install command

esc:
There was a missing double semi-colon to denote a command which meant that the
backticks were being swallowed. Also, replace the backticks in favour of '$( )'
since that is the modern and recommended way.


jaraco:
Thanks!

PR #220, opened 2016-05-09
Added License info to PKG-INFO

rickard-von-essen-iz:
Before this change:

``` shell
$ pip show keyring

---
Metadata-Version: 2.0
Name: keyring
Version: 9.0
Summary: Store and access your passwords safely.
Home-page: https://github.com/jaraco/keyring
Author: Jason R. Coombs
Author-email: jaraco@jaraco.com
Installer: pip
License: UNKNOWN
[...]
```

With the change _License_ will be _MIT_

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,7 @@ setup_params = dict(
     description="Store and access your passwords safely.",
     long_description=long_description,
     url="https://github.com/jaraco/keyring",
+    license = "MIT",
     packages=setuptools.find_packages(),
     include_package_data=True,
     install_requires=[


mitya57:
I want to note that according to classifiers the license is double (PSF or MIT).

jaraco:
I'm -1 on this change. According to [the authoritative docs](https://docs.python.org/3.5/distutils/setupscript.html#additional-meta-data), "The license field is a text indicating the license covering the package where the license is not a selection from the “License” Trove classifiers." The Trove classifiers are a more robust way to describe a license. The bug lies with pip. Adding the field in the improper place here will only serve to mask the bug in pip and perpetuate the misconception that the 'license' field is the best place to describe the license(s) of the package.

rickard-von-essen-iz:
OK, make sense. Even if it's a bit annoying that pip don't handle this in a better way.

Issue #83, opened 2015-02-24
KWallet keyring should store service name in a folder rather than in username


jaraco:
In pull request `#21` (no longer available since migration from Bitbucket), Maciej demonstrates how it would be possible to have the KWallet keyring backend store the service name in its own unique field using the 'folder' concept of KWallet.

---
- Bitbucket: https://bitbucket.org/kang/python-keyring-lib/issue/83
- Originally reported by: Jason R. Coombs
- Originally created at: 2013-01-08T23:16:04.699

jaraco:
I agree we should do this, but as I mentioned in the pull request, we should be careful about backward compatibility. In the past, Keyring has automatically migrated a user's data from older versions (usually a few minor releases past) before removing the migration support. I don't have any way to reach out to the users to ask what the impact of a manual migration or no migration would be, so I'm stuck recommending that we build an automatic migration for this feature as well. It's okay if the migration is not reversible, as long as the change is clearly communicated.

Maciej, can you work on a patch that only addresses this aspect and that performs automatic migration? Take a look at [the file location migration](https://bitbucket.org/kang/python-keyring-lib/src/fff4872e0b6be02ee2c7a6802d8f2c7129d210aa/keyring/backend.py?at=0.8#cl-290) for an example of something that worked.

Note that a migration will only be possible if KWallet facilitates enumerating all passwords, something the existing keyring doesn't yet demonstrate.

---

Original comment by: Jason R. Coombs

jaraco:
I'll work on that

---

Original comment by: Maciej Sitarz

jaraco:
I started from writing tests and I'm not sure if it's doable, I mean the migration.

From already existing tests I assume that username, password and service name allow Unicode, whitespace and punctuation characters.

```
#!python

DIFFICULT_CHARS = string.whitespace + string.punctuation
UNICODE_CHARS = escape.u("""κόσμεНа берегу пустынных волнSîne klâwen durh die
wolken sint geslagen, er stîget ûf mit grôzer kraft""")
password = random_string(20, DIFFICULT_CHARS)
username = random_string(20, DIFFICULT_CHARS)
service = random_string(20, DIFFICULT_CHARS)

```

string.punctuation contains "@" which is used as a username/service separator.
I should probably exclude "@" for service name in my test, but knowing this I think the automatic migration won't be successful in 100% cases.

---

Original comment by: Maciej Sitarz

jaraco:
I attached first version of a migration patch.

[...]

PR #229, opened 2016-06-14
KWallet keyring should store service name in a folder name rather than in username

maciex:
This is a continuation of work from issue #83.

I rewrote all from scratch. Since last time KWallet backend was rewritten to use DBus, so most of the changes was not mergeable.

Please review.

### DIFF
--- a/keyring/tests/test_backend.py
+++ b/keyring/tests/test_backend.py
@@ -14,9 +14,6 @@ from keyring.util import escape
 from .util import random_string
 from keyring import errors
 
-# Remove '@' from service name as this is not supported in service names
-# '@' will cause troubles during migration of kwallet entries
-DIFFICULT_CHARS = string.whitespace + string.punctuation.replace('@', '')
 # unicode only characters
 # Sourced from The Quick Brown Fox... Pangrams
 # http://www.columbia.edu/~fdc/utf8/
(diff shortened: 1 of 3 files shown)


maciex:
Few assumptions I made:
1. Service name is detected from old username, it takes a string after last '@'
2. '@' is not allowed in tests (removed from DIFFICULT_CHARS)
3. Converted entries are removed
4. Empty 'Python' folder is removed

keyring/tests/test_backend.py
```diff
@@ -14,7 +14,9 @@
 from .util import random_string
 from keyring import errors
 
-DIFFICULT_CHARS = string.whitespace + string.punctuation
```
jaraco, line 17: It's not reasonable to lower the bar of acceptability for all backends just to support a migration. If you want to override this value just for the kwallet backend, that might be reasonable, but even better would be to find a way to encode these characters such that they're not given special treatment w.r.t. other backends.

I apologize I haven't read the rest of the patch, so I don't have a good suggestion to address this concern.

jaraco:
Thanks for working on this. I hope @mitya57 can also review and give his approval or recommendations.

keyring/tests/test_backend.py
maciex, line 17: Sure, you're right. I just went the easiest path.

But I'm not sure what do you mean by "even better would be to find a way to encode these characters such that they're not given special treatment w.r.t. other backends".

I thought about overwriting that property only for migration test, would it be ok?

maciex, line 17: I moved DIFFICULT_CHARS into BackendBasicTests. '@' will be removed only for KWallet tests.

PR #232, opened 2016-06-23
require dbus-python and secretstorage on linux

reece:
This PR installs dbus-python and secretstorage on linux.

I test it with:

```
pip install -e git+https://github.com/reece/keyring@231-require-secretstorage-on-linux#egg=keyring
```

I did _not_ test it on other platforms, which should probably be done just to make sure it didn't foul up non-linux platforms. (Obviously, not expected.)

I intentionally left the current docs untouched because they might be referenced by folks on previous versions for the near future.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,12 @@ test_requirements = [
     'pytest>=2.8',
 ]
 
+linux_requirements = [
+    # dbus-python is a declared req only by some versions of secretstorage
+    "dbus-python",
+    "secretstorage"
+]
+
 setup_params = dict(
     name='keyring',
     use_scm_version=True,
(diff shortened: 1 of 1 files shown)


mitya57:
From the Travis logs:

```
error in keyring setup command: Invalid environment marker: platform_system=="Linux"
```

jaraco:
`platform_system` is an environment marker from PEP 508, added in Setuptools 20.2 and pip 8.1.2. Looks like it's probably too soon to use it. I suggest `sys_platform == "linux" or sys_platform == "linux2"` as the marker for now. Sorry I didn't catch this sooner.

PR #239, opened 2016-07-12
Update URL for pytest-runner

n8henrie:
Like keyring, it appears that pytest-runner now lives at GitHub instead of Bitbucket.


jaraco:
Thanks for the contrib.

PR #248, opened 2016-11-02
SecretStorage: add an option to specify custom collection path, and use that in tests

mitya57:
A session collection is temporary and is stored in the server process memory. Using it allows us to run tests when the default collection does not exist.

I use a similar patch in Debian packaging, and it allows us to run the tests with Secret Service backend in an automated manner, see [Debian logs](https://ci.debian.net/packages/p/python-keyring/unstable/amd64/) or [Ubuntu logs](http://autopkgtest.ubuntu.com/packages/python-keyring) (which have five different architectures).

### DIFF
--- a/keyring/backends/SecretService.py
+++ b/keyring/backends/SecretService.py
@@ -33,10 +33,16 @@ class Keyring(KeyringBackend):
                 "Unable to initialize SecretService: %s" % e)
         return 5
 
-    def get_default_collection(self):
+    def get_preferred_collection(self):
+        """If self.preferred_collection contains a D-Bus path, the collection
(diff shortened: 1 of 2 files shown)


mitya57:
@jaraco Do you mind if I merge this myself?

mitya57:
@mathstuf I remember you were trying to run keyring tests with your own Secret Service daemon. Does it support the `/org/freedesktop/secrets/collection/session` collection? Are you fine with this change?

jaraco:
Sorry I didn't see this sooner. Do feel free to merge it. Also, if you add a change to the published changes file and tag the release, it will get released automatically.

PR #250, opened 2016-11-20
Support both kwallet and kwallet5 dcop interfaces

frispete:
Hi, first of all, thank you for this great package. 

I had been working on a PyQt3 to PyQt5 transition of an inhouse app, and was in need of a kwallet interface. Unfortunately, I need to support both kwallet variants. The attached PR does exactly that, in a transparent way, and answers some questions in the doc string, I asked myself, when started experimenting with keyring.

Hopefully, the PR is acceptable for you. It's carefully designed to just adding what was needed.
One thing, I'm not 100% confident is, that it is using the dcop ProxyObject over the runtime of of keyring, expecting, that the dcop session will survive over the full keyring session.

### DIFF
--- a/keyring/backends/kwallet.py
+++ b/keyring/backends/kwallet.py
@@ -12,9 +12,51 @@ except ImportError:
 
 
 class DBusKeyring(KeyringBackend):
-    """KDE KWallet via D-Bus"""
+    """KDE KWallet via D-Bus
+       find out, which kwallet is in use with the wallet class attribute:
+
+       --> import keyring
+       --> kr = keyring.get_keyring()
+       --> kr
+       <keyring.backends.kwallet.DBusKeyring object at 0x7fac597492b0>
+
+       kwallet
+       --> kr.wallet.requested_bus_name
+       'org.kde.kwalletd'
+       --> kr.wallet.object_path
+       '/modules/kwalletd'
+
+       kwallet5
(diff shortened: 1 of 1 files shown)


frispete:
Follow up patch tries to deal with invalidated dbus sessions, as well as changing kwallets.

jaraco:
Hey Frispete. Glad to have you bringing contributions. This change seems reasonable and acceptable, though I'll probably re-write it slightly to implement it as two different backends. I'm not all that familiar with the Linux backends, but I do have a question - have you tried the SecretService backend? My understanding is it is has broader support and might work in both of the cases you're describing.

mitya57:
KWallet (still) does not support Secret Service :-(

jaraco:
Okay. That explains it. Thanks @mitya57.

jaraco:
@frispete: I've extracted a separate backend for each mode. Can you test with the latest head and let me know if it works in both environments?

mitya57:
Now that we have this, you can probably drop the old KWallet backend from keyrings.alt.

frispete:
Hi Jason, glad you liked it. Thanks for merging! Much appreciated. This way, keyring is able to support the majority of KDE installations out there.

jaraco:
> drop the old KWallet backend from keyrings.alt

Good idea.

PR #251, opened 2016-11-20
Reflect the KDE evolution in the class hierarchy, but keep the priority

frispete:
Get this in line with the usual KDE nomenclature, that adds a 5 for KDE Framework 5 elements, similar to the bus names and object paths. This way, I find it less confusing. Anyway, the priority order is kept. 
While at it, remove unused code.

I checked your changes and this one too for both environments, and it acts as advertised.
I love this code more than ever before... ;) Okay, okay, I'm in touch with it for a day now... :)

Cheers,
Pete

### DIFF
--- a/keyring/backends/kwallet.py
+++ b/keyring/backends/kwallet.py
@@ -13,27 +13,12 @@ except ImportError:
 
 class DBusKeyring(KeyringBackend):
     """
-    KDE KWallet 5 via D-Bus
+    KDE KWallet 4 via D-Bus
     """
 
     appid = 'Python program'
-    wallet = None
-    bus_name = 'org.kde.kwalletd5'
-    object_path = '/modules/kwalletd5'
-
-    @classmethod
-    def _select_wallet(cls, bus):
-        if cls.wallet is not None:
-            return cls.wallet
-
-        for bus_name, object_path in cls._wallet_objects:
-            try:
-                proxy = bus.get_object(bus_name, object_path)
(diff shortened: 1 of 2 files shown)


mitya57:
One day in a few years we will want to remove the KDE 4 variant as obsolete, and will have to revert this change…

frispete:
@mitya57: If I'm still alive and kicking at this point, I will do that with pleasure.

Isn't that a real joy with code, that avoids upsetting its readers? keyring is a perfect example of nicely organized code, that even allows unenlightened average people like me to contribute in useful ways.

mitya57:
@frispete Your work is welcome, and I did not downvote your pull request, it was just a side note.

frispete:
@mitya57 Didn't understood this as such, but wanted to use the opportunity to spread some serious compliments ;). Be assured, that I'm used to the full range of projects fitnesses, and this one is a good example for the positive side of things: codebase, responsiveness, kindness, tolerance, cooperation...

frispete:
Hence, it deserves support!

jaraco:
Thanks for all the kind words.

Unfortunately, I'm to decline this change, mainly because it introduces a backward-incompatibility for any users configuring the keyring by name, but also because it entangles the legacy behavior with the preferred, modern behavior. As currently written, when that support is no longer needed, the legacy classes can simply be removed and everything else remains the same. I will remove the extra unused method, though. Thanks for catching that.

PR #252, opened 2016-11-21
Guess, you will want to do this somehow

frispete:


mitya57:
Thanks! Indeed, the previous version did not work.

I hope @jaraco won't mind that I merged this :)

PR #255, opened 2016-12-04
Add a keyring backend name property

### LINKED ISSUE #253: backend discovery by name
Hi, what's the preferred method of backend discovery?

I want to supply a GUI selector for keyrings to my program, because I noticed, that due to testing and fixing some SecretService build issue here, the keyring in use moved from KWallet to Secretservice behind the scenes without any further notice. (Well, it may asked for credentials one additional time, but because I'm debugging all this, so it's gone unnoticed as well).

Here's a selection of what I found so far.

Using the object itself:
```
--> for kr in keyring.backend.get_all_keyring(): print(kr)
... 
<keyring.backends.kwallet.DBusKeyringKWallet4 object at 0x7f03eb60fa90>
<keyring.backends.SecretService.Keyring object at 0x7f03eb60f438>
<keyring.backends.fail.Keyring object at 0x7f03eb60f9b0>
```
Fiddling with object names is no fun, of course.

`__class__.__name__`:
```
--> for kr in keyring.backend.get_all_keyring(): print(kr.__class__.__name__)
... 
DBusKeyringKWallet4
Keyring
Keyring
```
my favorite, but not suitable for obvious reasons....

`__module__`:
```
--> for kr in keyring.backend.get_all_keyring(): print(kr.__module__)
... 
keyring.backends.kwallet
keyring.backends.SecretService
keyring.backends.fail
```
That would be nice, but fails to differenciate the kwallets..

Hm, boils down to to the object so far. Not that nice. 

Am I missing something obvious?

Since the kwallet thing isn't released, we could throw it into another module (kwallet4), and have `__module__` work well, but I haven't looked into keyrings.alt, if there are multiple backend classes in a single module, that need similar differentiation. 

I would love to prepare a related patch, if this approach is acceptable.

frispete:
Hi Jason, here we go in order to fix #253..

### DIFF
--- a/keyring/tests/backends/test_SecretService.py
+++ b/keyring/tests/backends/test_SecretService.py
@@ -12,7 +12,9 @@ class SecretServiceKeyringTestCase(BackendBasicTests, unittest.TestCase):
     def init_keyring(self):
         print("Testing SecretServiceKeyring; the following "
             "password prompts are for this keyring")
-        return SecretService.Keyring()
+        keyring = SecretService.Keyring()
+        keyring.preferred_collection = '/org/freedesktop/secrets/collection/session'
+        return keyring
 
 class SecretServiceKeyringUnitTests(unittest.TestCase):
     def test_supported_no_secretstorage(self):


frispete:
Hmm, travis bails out because of OS-X issues.

I noticed something related to OS-X as well, although I'm running linux: 
```
================================================= test session starts ==================================================
platform linux -- Python 3.4.5, pytest-3.0.4, py-1.4.31, pluggy-0.4.0
benchmark: 3.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=5.00us max_time=1.00s calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /home/hp/src/GIT/keyring, inifile: pytest.ini
plugins: cov-2.4.0, benchmark-3.0.0
collected 52 items / 1 errors 

======================================================== ERRORS ========================================================
____________________________________ ERROR collecting keyring/backends/_OS_X_API.py ____________________________________
backends/_OS_X_API.py:12: in <module>
    _sec = ctypes.CDLL(fw(name='Security'))
/usr/lib64/python3.4/ctypes/__init__.py:351: in __init__
    self._handle = _dlopen(self._name, mode)
E   OSError: /System/Library/Frameworks/Security.framework/Versions/A/Security: cannot open shared object file: No such file or directory
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
=============================================== 1 error in 1.01 seconds ================================================
```
Do you have any ideas, what's going wrong here? It shouldn't attempt to load some OS-X libs at all!

mitya57:
The `_OS_X_API` module should not be imported unless the OS is macOS.

Also, please rebase your changes against the latest master branch so that it’s easier to review them (actually rebase, not what you did).

PR #256, opened 2016-12-04
Add a keyring backend name property

frispete:
Here we go again to solve #253


jaraco:
I've gone ahead and merged this so I can work out the issues on Mac.

frispete:
Great, thanks.

PR #259, opened 2017-01-07
Allow to set a custom application attibute for SecretService backend

frispete:
This is handled similar to kwallet appid


frispete:
Should I close this PR, and create a new one - I wasn't aware, that this is still open?

mitya57:
This makes sense to me. A similar thing can be also done in the old GNOME backend which [also sets `application` attribute](https://github.com/jaraco/keyrings.alt/blob/master/keyrings/alt/Gnome.py#L79).

jaraco:
Rebased and merged as d4dc082.

PR #264, opened 2017-02-27
Add hook for supplying password to devpi-client.

jaraco:
Attempts to implement interface proposed in https://bitbucket.org/hpk42/devpi/pull-requests/397.

@fschulze, does this look like it would do what you expect? Would a `None` return from `devpiclient_get_password` defer to another plugin?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+10.3
+----
+
+* #264: Implement devpi hook for supplying a password when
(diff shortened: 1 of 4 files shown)


fschulze:
Looks correct and I used it to upload devpi-client 2.8.0.dev0 to https://devpi.net/fschulze/dev/

Yes, returning None defers to the next plugin. Actually the password prompt is implemented as the last tried plugin.

PR #277, opened 2017-05-28
Link to keyring_jeepney backend

takluyver:
I wrote a pure Python DBus interface, which avoids some of the installation hassle with depending on the system dbus libraries.

This is not super robust so far - e.g. it assumes that the default keyring is already unlocked, which I think is typically true but not always. But I have checked that I can store, retrieve and delete a password, at least on my own machine.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -154,6 +154,9 @@ use-cases. Simply install them to make them available:
 - `keyrings.alt <https://pypi.org/project/keyrings.alt>`_ - "alternate",
   less common backends, originally part of the core package, but now
   available for opt-in.
+- `keyring_jeepney <https://pypi.python.org/pypi/keyring_jeepney>`__ - a
+  pure Python backend using the secret service DBus API for desktop
+  Linux.
 
 Write your own keyring backend
 ==============================


README.rst
```diff
@@ -154,6 +154,9 @@ use-cases. Simply install them to make them available:
 - `keyrings.alt <https://pypi.org/project/keyrings.alt>`_ - "alternate",
   less common backends, originally part of the core package, but now
   available for opt-in.
+- `keyring_jeepney <https://pypi.python.org/pypi/keyring_jeepney>`__ - a
```
mitya57, line 157: One trailing underscore please, not two.

mitya57 (changes requested):

mitya57:
This looks very interesting, as this backend may solve many issues with installing python-keyring (there would be no need to compile dbus-python from source or use distro packages anymore).

I was considering porting my SecretStorage library away from python-dbus. Do you think your jeepney library is mature enough (or will be soon) so that I could use it?

mitya57 (approved):

README.rst
mitya57: Nevermind, I just learned that two underscores is also possible, and it means anonymous target.

takluyver: Yep, I'm in the habit of using two underscores to avoid Sphinx giving me warnings if there are two links with the same text. That's not going to be a problem here, though.

takluyver:
Thanks, I'm glad you're interested. :-)

I think this is the first non-example code using Jeepney. I'm fairly happy with how well the DBus communication seems to work, but not yet 100% convinced of the API design. I've chosen the Sans-IO design for the core so that the serialisation & parsing machinery can be used with different event loops (asyncio, tornado, different GUI event loops, synchronous IO...). But I'd like to reduce the code needed in each integration module to handle things like connecting to the bus and matching replies to method call messages. I'm also not sure about some of the high-level wrapper APIs.

There are a couple of limitations I'm aware of:
- It can only use Unix domain sockets. There are a variety of transports specified for DBus, including TCP sockets, but I don't know how commonly they come up - my Ubuntu and Fedora machines both use Unix sockets.
- It does not handle the transfer of file descriptors, which may be sent out-of-band over a Unix socket. I'm not sure how to make this possible when the stream parsing is separated from IO.

Jeepney is lower level than dbus-python; it works at the level of sending and receiving messages, rather than proxy objects with methods to call. One thing that I've considered is a bindings generator, which would use the XML introspection data available over dbus to generate classes which send and receive the necessary messages.

takluyver:
Oh, and there are no docs as yet, but that can be remedied.

jaraco:
Love it. Thanks for jeepney.

takluyver:
Thanks :-)

takluyver:
@mitya57 I should also have mentioned that Jeepney currently requires Python 3.5+, like most things that I've started as a side project. I might be talked into backporting it, but I'm not really enthusiastic about the notion.

mitya57:
All limitations sound like not critical for me.

If I manage to port Secret Storage to Jeepney (when I have time for that), I will maintain two branches: 2.x using dbus-python and 3.x using Jeepney.

takluyver:
Cool :-). If you do try it, let me know how you get on!

PR #289, opened 2017-10-19
Update cli.py

earonesty:
https://github.com/jaraco/keyring/issues/287

### DIFF
--- a/keyring/cli.py
+++ b/keyring/cli.py
@@ -7,8 +7,9 @@ import getpass
 from optparse import OptionParser
 import sys
 
-from . import get_keyring, set_keyring, get_password, set_password, delete_password
 from . import core
+from . import backend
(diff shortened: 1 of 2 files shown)


keyring/cli.py
```diff
@@ -21,6 +21,9 @@ def __init__(self):
         self.parser.add_option("-b", "--keyring-backend",
                                dest="keyring_backend", default=None,
                                help="Name of the keyring backend")
+        self.parser.add_option("-l", "--list-backends",
+                               dest="list_backends", action="store_true",
+                               help="Name of the keyring backend")
```
mitya57: The help text is copy-pasted from the previous entry, please update it.

mitya57 (changes requested):
How am I supposed to use it?

Just running `python3 -m keyring --list-backends` I get the help message because `len(args) == 0` and the code on line 36 is triggered.

PR #292, opened 2017-11-09
Set the appid for SecretService & KWallet to something meaningful.

jonathanverner:
On KDE, e.g., the appid is displayed by the KWallet when it prompts the user to open a wallet.  So someting more meaningful than 'Python program' should be sent as the appid.

### DIFF
--- a/keyring/backends/SecretService.py
+++ b/keyring/backends/SecretService.py
@@ -1,5 +1,8 @@
 import logging
 
+import sys
+import os
+
 from ..util import properties
 from ..backend import KeyringBackend
 from ..errors import (InitError, PasswordDeleteError,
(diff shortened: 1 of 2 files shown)


mitya57:
Will it work with the already saved items from older python-keyring versions?

jonathanverner:
In KWallet, the `appid` attribute does not influence where the passwords are stored. It is only used to decide whether to allow the app access or not, so it should work with items saved from older versions. 

For SecretService, as currently implemented, it should also work because the [code](https://github.com/jonathanverner/keyring/blob/940377db8224b3437c8ab70b60964ca7a967340d/keyring/backends/SecretService.py#L62) in `get_password ` restricts its search for the password by `service` and `username` only (i.e. it disregards the `application` attribute).

However, it might be a good idea, in the future, to make SecretService also restrict its search by `application` (so that it does not conflict with other applications). If that is done, one could first try restricting the search to the new `appid` and if that fails, either fall back to `python-keyring` or to the current method (i.e. disregard the `application` attribute).

mitya57:
OK, I don't mind then.

The Travis failure looks like a problem with macOS test runners, they cannot resolve `github.com` for some reason.

PR #304, opened 2018-02-15
More specific DE hints

frispete:


jaraco:
I cherry-picked this commit as 6ec3fdd.

PR #309, opened 2018-03-05
Direct module loading

yanone:
The previously cryptic module loading procedure prevented py2app from finding and adding the various backend files, producing import errors in the compiled app.
Since you've manually specified the backend names in _load_backends() anyway, why not just import them like this?
Now the compiled app works.

### DIFF
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -127,19 +127,12 @@ class NullCrypter(Crypter):
         return value
 
 
-def _load_backend(name):
-    "Load a backend by name"
-    package = backends.__package__ or backends.__name__
-    mod = importlib.import_module('.' + name, package)
-    # invoke __name__ on each module to ensure it's loaded in demand-import
-    # environments
-    mod.__name__
-
-
 def _load_backends():
     "ensure that native keyring backends are loaded"
-    backends = 'kwallet', 'OS_X', 'SecretService', 'Windows'
-    list(map(_load_backend, backends))
+    import keyring.backends.OS_X
+    import keyring.backends.Windows
+    import keyring.backends.kwallet
+    import keyring.backends.SecretService
     _load_plugins()
 
 


yanone:
Sorry, I just saw the build checks.
I'm on Python 2.7.14. I haven't arrived at Python 3 yet.

jaraco:
Thanks for putting together a PR. Your approach is more direct and clear.

I do generally avoid `import` statements except at the top of the file. Usually when I have execution-time imports, I use `__import__` or `importlib.import_module` (if available). But that approach isn't recognized by py2app either, so that wouldn't solve the problem.

But more importantly, I've been for some time thinking that these backends should be presented through the entry-points mechanism, such that there's exactly one way backends are discovered, which would eliminate this code altogether.

The underlying problem here, and it applies to other systems that repackage applications (pyinstaller, py2exe, ...), is that py2app is not using established means of discovering packages and resolving dependencies, but is instead trying to [infer the dependencies based on imports](https://py2app.readthedocs.io/en/latest/implementation.html#depdency-resolution-via-modulegraph). Better than py2exe requiring all projects to write their code in a certain way would be to instead either:

- use established mechanisms to discover modules (the same mechanisms that distutils/setuptools does when it builds/installs a package). Of course, that requires access to the source.
- give each package a hook they can supply to declare their hidden dependencies [as pyinstaller does](https://github.com/jaraco/keyring/blob/master/hook-keyring.backend.py). In fact, there's nothing preventing py2app from using the convention created by pyinstaller to discover hidden packages. If py2app adopted this mechanism, it would no longer need to solicit [recipes](https://py2app.readthedocs.io/en/latest/recipes.html) for each and every package, but instead each package could declare its hidden imports (without affecting its preferred behavior).

So I guess it boils down to, while this change makes sense in isolation, it's difficult to accept the change due to the other constraints it applies.

Regarding the build check failures, I don't think they're due to your changes, but are probably due to Homebrew switching to Python 3... which seems to have caused pip to go missing (by default) on macOS.

jaraco:
I filed #310 to capture the previously undocumented intention I have to use the entry points for all backends.

jaraco:
I closed and re-opened to allow the tests to re-run now that master passes on macOS.

keyring/backend.py
```diff
@@ -127,19 +127,12 @@ def decrypt(self, value):
         return value
 
 
-def _load_backend(name):
-    "Load a backend by name"
-    package = backends.__package__ or backends.__name__
-    mod = importlib.import_module('.' + name, package)
-    # invoke __name__ on each module to ensure it's loaded in demand-import
```
jaraco, line 134: This important behavior is flagged with a comment to prevent it from being inadvertently optimized away, which this change does. So unfortunately, this change will break keyring in environments where demand-import is used (Mercurial in particular).

jaraco (commented):
As you can see in #310 and the associated commit I've drafted, this proposed change runs in conflict with other intentions of the project. Moreover, because it will likely break keyring in demand-import environments, it's not viable.

Incidentally, the tests are also failing now too because I've added a lint checker to the project, which is the sole reason the tests are currently failing. (no blame here because those checks weren't in place when you submitted this PR).

Thank you for the PR, but for the reasons outlined above, I don't think the project can accept this change at this time.

PR #313, opened 2018-03-29
Document the exceptions raised when library fails to get or set passw…

maxking:
Document the exceptions raised.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -292,6 +292,17 @@ simple ``str`` in the default encoding as they will be implicitly
 decoded to text. Some backends may accept ``bytes`` for these parameters,
 but such usage is discouraged.
 
+
+Exceptions
+==========
+
+The keyring lib raises following exceptions:
+
+* ``keyring.errors.KeyringError``: Base Error class for all exceptions in keyring lib.
+* ``keyring.errors.InitError``: Raised when the keyring can't be initialized.
+* ``keyring.errors.PasswordSetError``: Raise when password can't be set in the keyring.
+* ``keyring.errors.PasswordDeleteError``: Raised when the password can't be deleted in the keyring.
+
 ------------
 Get involved
 ------------


jaraco:
This seems reasonable to me, but it does enhance the expectation one will have on the library. Does this behavior seem correct to you?

maxking:
Ideally, I think these errors should be a part of the the public API and should be documented. I had a hard time trying to figure out when my system keyring was behaving badly and had to catch generic `Exception` to make sure that my app doesn't blow up.

I think exposing errors as public API would make sense.

jaraco:
I agree, but I guess my question was - does the API seem like a good one to lock in on, or would you expect something different? If the latter, we should transition to that before locking in.

README.rst
```diff
@@ -292,6 +292,17 @@ simple ``str`` in the default encoding as they will be implicitly
...
 but such usage is discouraged.
 
+
+Exceptions
+==========
+
+The keyring lib raises following exceptions:
+
+* ``keyring.errors.KeyringError``: Base Error class for all exceptions in keyring lib.
+* ``keyring.errors.InitError``: Raised when the password can't be initialized.
```
maxking: Looks like a typo here, it should be "the keyring can't be initialized" I suppose.

maxking:
I think the API looks good, but it should be exposed through top level import.

jaraco:
> I think the API looks good, but it should be exposed through top level import.

I could go either way on this. On one hand, [There should be one-- and preferably only one --obvious way to do it.](https://www.python.org/dev/peps/pep-0020/). On the other hand, [practicality beats purity](https://www.python.org/dev/peps/pep-0020/). Given the apparently rare situation that exception trapping arises, I'm inclined to lean toward not duplicating the names and documenting them as you have.

maxking:
Cool, I can fix the typo and push the changes.

frispete:
Guess, the hard part will be catching all exceptions from the underlying libs and map them to this scheme. Don't know about keyring itself, but keyrings.cryptfile as well as a couple of (deprecated alt backends) doesn't do that well, atm.

PR #315, opened 2018-04-07
Make KWallet keyring have higher priority than Secret Service on KDE

### LINKED ISSUE #273: In KDE Plasma, keyring uses gnome-keyring by default
In a relatively fresh install of Fedora 25 with KDE Plasma as the DE, seems keyring prefers to use gnome-keyring (libgnome-keyring) by default.
Getting rid of gnome-keyring doesn't solve the problem, while erasing libgnome-keyring is not feasible as it is a dependency of many packages.
Any chance to add some logic to detect the desktop environment and use kwallet if running under plasma?

mitya57:
Fixes #273.


frispete:
Great, @mitya57, much appreciated.

PR #322, opened 2018-04-26
bugfix: AttributeError: 'dict' object has no attribute 'unichr'

micahculpepper:
Hello,

My users (Python 2) have started getting this crash after upgrading yesterday to keyring 12.1.0 (from 10.4.0):

```
  File ".../lib/python2.7/site-packages/keyring/util/escape.py", line 27, in _unichr
    return __builtins__.unichr(c)
AttributeError: 'dict' object has no attribute 'unichr'
```

It seems like 09c9008d7c71a2fb83264e9eeefa4f5c0daa9ae8 broke something. I confirmed that this PR fixes it for me.

I wish I knew why _exactly_ this is a problem, because the new form seems correct, and if I open up an interpreter and do a `dir(__builtins__)`, `unichr` is in there. However, if I throw a print statement before the `def _unichr`, I see that somehow, in that module, `__builtins__` only contains this stuff:

```
['__class__', '__cmp__', '__contains__', '__delattr__', '__delitem__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getitem__', '__gt__', '__hash__', '__init__', '__iter__', '__le__', '__len__', '__lt__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__setitem__', '__sizeof__', '__str__', '__subclasshook__', 'clear', 'copy', 'fromkeys', 'get', 'has_key', 'items', 'iteritems', 'iterkeys', 'itervalues', 'keys', 'pop', 'popitem', 'setdefault', 'update', 'values', 'viewitems', 'viewkeys', 'viewvalues']
```

and `unichr` is missing.

So somehow, something in my pile of scripts is altering `__builtins__`? Not sure. Could be related to my attempts to be Py2/3 compatible through `__future__` imports  and the `futurize` library. But in any case, if folks are shadowing `unichr`, there's probably a reason for it, and we should just accept the new name, right?

### DIFF
--- a/keyring/util/escape.py
+++ b/keyring/util/escape.py
@@ -23,7 +23,7 @@ else:
         return s.decode('utf-8')
 
     def _unichr(c):
-        return unichr(c)
+        return unichr(c)  # noqa: F821
 
 LEGAL_CHARS = (
     getattr(string, 'letters', None)  # Python 2
--- a/keyring/util/escape.py
+++ b/keyring/util/escape.py
@@ -23,7 +23,7 @@ else:
         return s.decode('utf-8')
 
     def _unichr(c):
-        return __builtins__.unichr(c)
+        return unichr(c)
 
 LEGAL_CHARS = (
     getattr(string, 'letters', None)  # Python 2


micahculpepper:
~~If I'm reading this right, the tests are failing on Py3. But the code shouldn't be executed on Py3?~~

mitya57:
It is not tests, it is flake8 complaining on Python 3. Try adding `# noqa` to the end of line to silence it.

mitya57 (approved):

micahculpepper:
Thanks for the approval @mitya57. What is the process for requesting a new release?

mitya57:
I think you should wait for @jaraco who is this repo's owner. He usually merges and releases very quickly :)

jaraco:
I'm not necessarily opposed to this change, though I am a little reluctant because it would be nice to uncover the _actual_ cause of the failure. Looking at [the docs](https://docs.python.org/2/library/__builtin__.html), I'm fairly sure the use of `__builtins__` was a bad move, not only because it's an implementation detail but also because its state is poorly specified (it can be either the builtins or the module's `__dict__`, which is what I think you're seeing). I think I chose it because the name was readily available without any linter concerns.

I would like to explore another technique for this, but at first blush, it's the best approach.

micahculpepper:
@jaraco Thanks for merging that in! I would also like to know what the actual cause is. I'm still not sure, but I got a little closer with a minimal example:

```
~ → cat mymodule.py
print __builtins__.unichr

 ~ → cat tester.py
import mymodule

 ~ → python2 mymodule.py
<built-in function unichr>

 ~ → python2 tester.py
  File "tester.py", line 1, in <module>
    import mymodule
  File "/Users/micah/mymodule.py", line 1, in <module>
    print __builtins__.unichr
AttributeError: 'dict' object has no attribute 'unichr'

~ → python2 -V
Python 2.7.14
```

So the value of `__builtins__` seems to change depending on whether or not you are running at the top level... or something. This is above my pay grade ;)

PR #328, opened 2018-05-15
Improve kwallet backend

### LINKED ISSUE #298: Reopening KWallets
Hi,

unfortunately, here's yet another issue with kwallets.. 😢 

This time, it's triggered by the following steps:
 * open `kwallet` keyring
   * if not yet opened, it asks for the `kwallet` password
   * if `kwallet` was opened already, it asks for confirmation of this app to access the wallet (unless you permitted access to this app generally before)
 * fetch passwords from `kwallet`
 * force close the wallet (via `kwalletmanager`) 
 * when fetching more passwords now, one would expect another request for permission, but `get_password()` simply fails (returning None)

I found a workaround, but it's **butt ugly**. If `get_password()` fails, I reset the backend handle to -1, and reinitialize the keyring. That forces the connection to re-establish, and requests either for the wallets password, or permission as expected. Hence, the `kwallet backend` is missing some **disconnection detection** to handle this situation gracefully. `connected()` does not fit this purpose...

Any idea anybody, how to approach this properly?

### LINKED ISSUE #327: Kwallet: Give meaningful error if the keyring has not been unlocked
Hi,

When i start the application, Kwallet prompts for the password, if i dont enter the password, `get_password()` returns after 20 seconds with `None` without any error.

Per docs this means there is no password stored in the keyring, which is wrong in this case. I need to know if the user didnt unlock his keyring, because then i can try again. Trying again makes no sense if there is no password in the keyring.

lovetox:
If connected() returns on all exceptions False, applications have no way of knowing if the password is not in the keyring or an error has happened.

Fixes #327 

The wallet could have been closed after a while. Check if the wallet is
open before using it

Fixes #298

### DIFF
--- a/keyring/backends/SecretService.py
+++ b/keyring/backends/SecretService.py
@@ -6,7 +6,7 @@ import os
 from ..util import properties
 from ..backend import KeyringBackend
 from ..errors import (InitError, PasswordDeleteError,
-                      ExceptionRaisedContext)
+                      ExceptionRaisedContext, KeyringLocked)
 
 try:
     import secretstorage
@@ -58,7 +58,7 @@ class Keyring(KeyringBackend):
         if collection.is_locked():
             collection.unlock()
             if collection.is_locked():  # User dismissed the prompt
-                raise InitError("Failed to unlock the collection!")
+                raise KeyringLocked("Failed to unlock the collection!")
         return collection
 
     def get_password(self, service, username):
@@ -71,7 +71,7 @@ class Keyring(KeyringBackend):
             if hasattr(item, 'unlock'):
                 item.unlock()
             if item.is_locked():  # User dismissed the prompt
-                raise InitError('Failed to unlock the item!')
+                raise KeyringLocked('Failed to unlock the item!')
             return item.get_secret().decode('utf-8')
 
     def set_password(self, service, username, password):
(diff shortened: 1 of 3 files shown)


lovetox:
I updated my MR with two more commits.

the last one `Raise KeyringLocked exception when appropriate` is something i would find useful, let me know what you think

jaraco:
These changes seem reasonable to me, but it does deviate from the existing paradigm, where a failure results in a general None response. Imagine a client that's written like:

```
password = keyring.get_password(system, user) or getpass.getpass()
```

Currently, if the user dismisses the dialog, get_password returns None and getpass gets invoked. In this now model, cancelling the dialog would cause an Exception to bubble through.

In the one real-world example I was considering [in setuptools upload](https://github.com/pypa/setuptools/blob/c13fcfd43f5ce16a2fb4930622293395a494004f/setuptools/command/upload.py#L25-L33), that wouldn't be a concern because all exceptions are trapped and transformed into None.

I also wonder to what extent the library should specify and other backends should also support this approach.

I'm happy to accept it and see how it's received.

jaraco (approved):

lovetox:
Thanks for looking over this.

If thats the paradigm then we should change it. If the library always returns None it withholds information that is vital for even really basic usecases.

Imagin some app that needs a password to login to a server. Then i need to show the user a dialog to enter the password, if there is no password in the keyring. But this dialog is totally wrong if the user dismisses the prompt. This would mean "I dont want to deal with this app now" and the app would have to interpret it as "We have no password, show a prompt to put in a new one". I think this is a very common scenario and with the lib returning in both cases None the developer can not differ between "no entry in the keyring" and "user does not want to deal with it now"

PR #332, opened 2018-06-13
Allow to set password from pipe

j-martin:
The usecase is to dump the output of a command into the keyring. For
example a random token is generated and must be stored.

### DIFF
--- a/keyring/cli.py
+++ b/keyring/cli.py
@@ -92,7 +92,7 @@ class CommandLineTool(object):
         if sys.stdin.isatty():
             return getpass.getpass(prompt)
         else:
-            return sys.stdin.readline().rstrip()
+            return sys.stdin.readline().rstrip('\n')
 
     def output_password(self, password):
         """Output the password to the user.
--- a/keyring/cli.py
+++ b/keyring/cli.py
@@ -89,8 +89,10 @@ class CommandLineTool(object):
 
         This mostly exists to ease the testing process.
         """
-
-        return getpass.getpass(prompt)
+        if sys.stdin.isatty():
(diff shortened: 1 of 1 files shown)


jaraco (commented):
I like this - definitely a good idea.

Few questions - 

1. Is this the pattern used by other command line interfaces? Can you share a reference?
2. Is `isatty` available on all platforms?
3. The stdin method only reads one line and strips any trailing whitespace. What if the password is multiline or contains trailing whitespace? My instinct is to remove one trailing newline if present, but otherwise accept the whole payload.
4. Does sys.stdin always yield text (unicode) on Python 3?

j-martin:
Hey Jason, 

1. I have not seen this pattern other than the internal tool I've been working on.
Basically, I got this shell script which generates a secret and I want to store in the keyring. Something like:

        $ openssl rand -base64 64 | tr -d "\n/+=" | keyring set application api-key

2. It's available on python 2.7+ (I have not checked older versions) and works on Windows (checked just now).
3. Make sense, I updated the PR.
4. Seemingly so, python3 will throw a UnicodeDecodeError exception. I think it is reasonable to throw an exception in this case.

        $ openssl rand 128 | python3 -c "import sys; print(sys.stdin.readline().rstrip('\n'))"

```
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe4 in position 0: invalid continuation byte
```

PR #346, opened 2018-10-05
Remove application attribute from stored passwords using SecretService

### LINKED ISSUE #345: appid field in SecretService keyring backend causes duplication issues.
When you save a password using this library, the password contains an 'application' field containing the app that set the password (python, or the name of the python script from which the function is called). Two otherwise identical passwords can be stored if they are different only by that field.

This can cause different scripts to create duplicates of the same password with no way of selecting which of the two passwords they want to retrieve with get_password. Get_password doesn't even use the 'application' field but will return the most recent password meeting the service/username criteria. delete_password() will delete the first one too.

pplatanias:
Fixes #345.

### DIFF
--- a/keyring/backends/SecretService.py
+++ b/keyring/backends/SecretService.py
@@ -1,8 +1,5 @@
 import logging
 
-import sys
-import os
-
 from ..util import properties
 from ..backend import KeyringBackend
 from ..errors import (InitError, PasswordDeleteError,
--- a/keyring/backends/SecretService.py
+++ b/keyring/backends/SecretService.py
@@ -22,7 +22,7 @@ log = logging.getLogger(__name__)
 
 class Keyring(KeyringBackend):
     """Secret Service Keyring"""
-    appid = os.path.basename(sys.argv[0]) or 'Python keyring library'
+    appid = 'Python keyring library'
 
     @properties.ClassProperty
     @classmethod


mitya57:
Note: this effectively reverts the SecretService part of #292 (by @jonathanverner). Indeed, when merging that request we did not think about potential regressions it would bring.

@jaraco What do you think? If you approve, then I will merge it.

pplatanias:
Seems like #292 was made mainly with KDWallet in mind. I'm not sure if the duplication issue exists there as well but Gnome Keyring doesn't seem to use to the application field in any popups.

jaraco:
@mitya57 Thanks for the analysis. I have no objections, but I'm mostly deferring to your judgment on this issue.

jaraco (approved):

jonathanverner:
Actually, I think that #345 should be solved in a way which would not also re-introduce #292... In particular, before saving a password, the script can explicitly set the application id (by setting the attribute `appid` on the keyring). This way it has control over where the password is written. Perhaps, it would make sense for the `get?password` method to take the appid into account when searching...

jaraco:
Reading through the issues again, I have to say, given the choice between #292 and #345, it's more important to fix #345. In important part of the design of keyring is that different applications should be able to share the same credentials... and client applications should have a consistent interface across all the backends.

My understanding was that the fix for #292 was meant to address the display of the app name and not the location of where the credential is stored, the location which should be dependent only on the 'system' and 'username'. If that's not the behavior, I believe we want to correct that behavior, and it sounds like this PR does that.

> before saving a password, the script can explicitly set the application id

I don't think that's generally possible without violating the aforementioned intended behavior. If the backend wishes to store passwords on a per-client basis, that sounds like a functionality that should not be honored through the default get_password/set_password interface.

@jonathanverner can you see why it's surprising and undesirable for one backend to discriminate based on `sys.argv[0]` but for other backends to share credentials across apps?

Can you clarify my understanding or help me understand why keyring should support that behavior?

PR #351, opened 2018-10-27
Fixes #350 Add get_username_and_password API

### LINKED ISSUE #350: Add `get_username_and_password` API
There are some scenarios where it is possible (and preferred) to retrieve the username for a service from just the system, but currently keyring has no ability to return a username that has not been provided by the caller.

I propose to add a new API matching `get_password`'s signature, but named `get_username_and_password` ad returning `(username, password)`, where `username` _might_ match the provided username. The ABC can include a default implementation of this that uses `get_password` and returns the provided username.

zooba:
Fixes #350 Add get_username_and_password API
Adds a new API to enable backends to return a username along with the password.
The entry points include fallback behavior for backends that do not implement the API.

### DIFF
--- a/keyring/__init__.py
+++ b/keyring/__init__.py
@@ -1,9 +1,9 @@
 from __future__ import absolute_import
 
 from .core import (set_keyring, get_keyring, set_password, get_password,
-                   delete_password)
+                   delete_password, get_credential)
 
 __all__ = (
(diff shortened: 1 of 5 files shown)


zooba:
I implemented this for Windows, because I can easily test it. Happy to do other platforms as well, but they're also good tests of the fallback

PR #353, opened 2018-10-28
Fixes #352 Allow all backends to participate

### LINKED ISSUE #352: Allow all backends to participate
Currently, only the highest priority backend (for the special definition of "priority") is used for getting and setting credentials.

I propose the default behavior (that is, before `set_keyring` is called) to allow all discovered backends to participate. My proposed semantics:

## init_backend
Instead of using `max` to select the best keyring in the unconfigured case, use a new backend that includes all those found/filtered by the call (`keyrings` local variable).

## get_password and get_credential
When getting a password, forward the call to each backend in order and return the first non-None result.

## set_password and delete_password
When setting a password, forward the call to each backend in order and return after the first that _does not_ raise `NotImplementedError`.

zooba:


jaraco:
@zooba Would you be willing to write up a brief about what motivated this change? I know we discussed it at the time, but it's since slipped my mind.

Issue #357, opened 2018-11-01
keyring 16.0.0 breaks keyrings.alt.file.PlaintextKeyring


basak:
Steps to reproduce:

Start a fresh Python 3 environment on Ubuntu Bionic. For example:

```
lxc launch ubuntu:bionic keyring
lxc exec keyring -- login -f ubuntu
sudo apt update && sudo apt -y install virtualenv
virtualenv --python=python3 keyring
. keyring/bin/activate
```
Install `keyring` 16.0.0 and `keyrings.alt`:

```
pip3 install keyring==16.0.0 keyrings.alt
```
(this uses `keyrings.alt` 3.1 which is currently the latest in PyPI)

Run the following:

```
python3 -c 'import keyring; print(keyring.get_keyring())'
```

Expected result:

```
keyrings.alt.file.PlaintextKeyring (priority: 0.5)
```

Actual result:

```
keyring.backends.fail.Keyring (priority: 0)
```

Downgrading `keyring` to 15.2.0 works around the problem:

```
$ pip3 install keyring==15.2.0
[...]
$ python3 -c 'import keyring; print(keyring.get_keyring())'
keyrings.alt.file.PlaintextKeyring (priority: 0.5)
```

mitya57:
I would say it is an issue in keyring itself, not in keyrings.alt.

For some reason the new [`_create_chained_backend` method](https://github.com/jaraco/keyring/blob/master/keyring/core.py#L86) ignores backends with priority < 1, such as PlaintextKeyring.

If I replace the check on line 91 with `priority > 0` then it almost works. I say almost because `print(keyring.get_keyring())` still fails: ChainerBackend has no priority, so its `__str__` method fails.

@jaraco Can you look at this please?

PR #358, opened 2018-11-05
Restore low-priority (0, 1) backends

jaraco:
In #357, we learned that low priority backends are now being excluded from use. That was not an intended change. I think I imagined that only the chainer would be affected, but because the chainer is now in place as the default mechanism for resolving backends, it had the effect of excluding the non-recommended backends.

This change restores support for non-recommended backends.

I think the main reason for adding the limit was to prevent the fail.Keyring from being included, so in this change, backends with a priority of 0 are still excluded from the chained backends.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+16.0.1
+------
+
+* #357: Once again allow all positive, non-zero priority
+  keyrings to participate.
+
 16.0.0
 ------
 
--- a/conftest.py
+++ b/conftest.py
@@ -8,3 +8,14 @@ if platform.system() != 'Darwin':
     collect_ignore.append('keyring/backends/_OS_X_API.py')
 
 collect_ignore.append('keyring/devpi_client.py')
+
+
+def pytest_configure():
+        workaround_sugar_issue_159()
+
+
+def workaround_sugar_issue_159():
+    "https://github.com/Frozenball/pytest-sugar/159"
+    import pytest_sugar
+    pytest_sugar.SugarTerminalReporter.pytest_runtest_logfinish = \
+        lambda self: None
--- a/keyring/backends/chainer.py
+++ b/keyring/backends/chainer.py
@@ -11,6 +11,14 @@ from ..backend import KeyringBackend
 
 
 class ChainerBackend(KeyringBackend):
+    """
+    >>> ChainerBackend(())
+    <keyring.backends.chainer.ChainerBackend object at ...>
+    """
+
+    priority = 0
+    viable = False
+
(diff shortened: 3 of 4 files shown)


mitya57:
printing the backend is still broken:
```
$ python3 -c 'import keyring; print(keyring.get_keyring())'
  File "/home/dmitry/upstream/keyring/keyring/backend.py", line 89, in __str__
    keyring_class.priority))
TypeError: must be real number, not function
```

zooba:
I'd suggest adding `priority = 0` to `ChainerBackend` as well, which should fix the `__str__` and also make sure we don't recursively chain backends.

jaraco:
Good catch!

jaraco:
@zooba I'm thinking about changing the design completely to this:

- extract the chainer backend to its own library.
- users install the chainer backend to opt in to the behavior.
- restore backend discovery to the prior behavior (highest priority is installed).
- chainer has a very high priority (10 maybe).
- chainer backend initializes itself with all viable keyrings (not itself).

This design would have better separation of concerns and give each user/environment more control over how it's used. Keyring might expose an `extra` so you could install `keyring[chained]` or `keyring[multi]`.

How does that approach strike you? How well does it satisfy your use-case?

jaraco:
As soon as I write that, I realize one possible downside - if keyring were to be adopted by stdlib, this behavior is yet another that's not available out of the box. I can't decide if that's an advantage or disadvantage.

jaraco:
Regardless of my comment above, it's probably worth getting a fix out for the issue, so I'll release a 16.0.1 with these changes and those proposed.

mitya57:
I like the proposed idea, it makes the behavior more predictive.

How high is the chance that keyring is adopted by stdlib? In this case some backends won’t be available anyway, because of missing dependencies.

zooba:
I think the chainer is better off in keyring itself, even if it's off by default. But I think as 2FA and password managers gain more appeal, we're more likely to want it on by default.

Only having a single backend forces the tools using it to implement multiple backends - see e.g. pip's handling of netrc and credentials in the URL, both of which could be easily handled by keyring backends.

In terms of putting it in the stdlib, I suspect it would be a requirement that multiple backends be supported. (Obviously I'd vote for that :) ) Alternatively, the setup should be changed so that a higher priority backend is provided with the next priority one, and it can chain manually if it so desires (but without having to parse configuration files itself).

Basically:
* trivial and obvious for a backend to break the chain (return something/raise)
* complex and non-obvious for a sole backend to correctly chain

Means it's better for the core library to handle chaining, and let the backends or configuration break it if they don't want chaining.

PR #360, opened 2018-11-13
Rewrite ChainerBackend as a normal, but preferred, backend

jaraco:


jaraco:
I want to align the chainer backend with the behavior of the other backends as much as possible.

I still have some reservations about this change:

- chainer is used in all cases, even when only one subordinate backend is available.
- NotImplementedError will be suppressed on set/delete.

PR #361, opened 2018-11-17
Implement Chainer as normal backend

jaraco:
In these changes, I revert all of the changes to the 'core' module from the original introduction of the ChainerBackend and instead implement the backend as any other discoverable backend, giving it the behaviors that give it priority when there are multiple backends suitable for chaining, but having it defer to other (by way of a 0 priority) when there are not chainable backends.

cc @zooba (FYI)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+16.1.0
+------
+
+* Refactor ChainerBackend, introduced in 16.0 to function
+  as any other backend, activating when relevant.
+
 16.0.2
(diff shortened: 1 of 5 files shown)


zooba:
@jaraco So to be clear, the default is still to chain all available backends? Or are we back to selecting only one backend and users have to both install a package _and_ modify a configuration file in order to get the chaining behavior?

jaraco:
Default is still to chain all available backends if there are at least two available.

PR #367, opened 2018-12-06
In ChainerBackend, honor any limit that might have been set during init_backend.

### LINKED ISSUE #366: Backends filtered by init_keyring aren't honored by chainer
In #365, I discovered that the intended interface of `keyring.core.init_backend` won't be honored due to the introduction of the `chainer` backend.

jaraco:
Fixes #366.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+17.1.0
+------
+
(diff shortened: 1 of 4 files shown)


jaraco:
cc @zooba FYI

Issue #371, opened 2019-01-11
Keyring tells user to use non-recommended backends


cowlinator:
In the case where no backend can be found, keyring errors with `RuntimeError: No recommended backend was available.  Install the keyrings.alt package if you want to use the non-recommended backends.`

Not only are `keyrings.alt` backends not recommended, they "may have security risks or other implications".

This, in spite of the fact that there are many valid 3rd party backends available.  

I appreciate the fact that this is documented in the README, but the error message should also state this fact or list a url where this fact can be found.

cowlinator:
Fixed with https://github.com/jaraco/keyring/pull/373

PR #373, opened 2019-01-11
Fix for issue #371:  Keyring no longer tells user to use non-recommended backend.

cowlinator:


cowlinator:
This is a fix for https://github.com/jaraco/keyring/issues/371

Issue #355, opened 2018-10-30
Obscure error when credential is longer than allowed


arkanoid87:
Using Windows backend, when supplying a password of a sufficient length, the following error is observed: `win32cred.CredWrite(credential, 0) -> (1783, 'CredWrite', 'The stub received bad data')'.`

arkanoid87:
Nevermind, solved.

If you encounter this, it is due to maximum password length in windows vault.

jaraco:
I'd still like to see this addressed, to trap the error and return a better one... or since the error message is obscure and doesn't talk about limitations, maybe just hard-code the limitation and return a nicer error message when the password length exceeds the limits.

I can't tell exactly what the limits are, but [these docs get close](https://docs.microsoft.com/en-us/windows/desktop/api/wincred/ns-wincred-_credentiala). I welcome contribution to this issue, especially from someone with some familiarity with CredentialVault.

@zooba If you know anybody who would like to help with this issue, please let them know. Thanks, sorry for the ping, and feel free to unsubscribe.

zooba:
Happy to be pinged :)

My reading of the docs is that the limit is always 512 bytes. So trapping the error and rewriting it when this limit is exceeded sounds fine. I wouldn't preempt the check, just in case the limit gets raised one day - rewriting a known error is safer.

(The "513 bytes" limit for username is to include a trailing null, which the blob does not require because the size is passed explicitly.)

ahqardah:
@jaraco we are facing the same issue on windows 10 and windows server 2019

```
File "{userpath}\appdata\local\continuum\miniconda3\envs\bonsai\lib\contextlib.py", line 130, in __exit__
    self.gen.throw(type, value, traceback)
  File "{userpath}\appdata\local\continuum\miniconda3\envs\bonsai\lib\site-packages\win32ctypes\pywin32\pywintypes.py", line 37, in pywin32error
    raise error(exception.winerror, exception.function, exception.strerror)
win32ctypes.pywin32.pywintypes.error: (1783, 'CredWrite', 'The stub received bad data').
```

 Are there any fixes for this?

jaraco:
I don't have a fix for this issue and can't even conceive of a suitable, proper fix. Contributions are welcome, but given the inherent limitations of the Windows Credential Store, I don't think you're going to find a solution that allows storing of these longer credentials. I'm open to ideas and contributions.

jaraco:
[...]

PR #374, opened 2019-01-12
Fix for #355.  

cowlinator:
win32cred error caused by password too long is handled such that a useful error message is printed (without destroying the original stacktrace)

### DIFF
--- a/keyring/backends/Windows.py
+++ b/keyring/backends/Windows.py
@@ -6,7 +6,7 @@ from ..py27compat import text_type
 from ..util import properties
 from ..backend import KeyringBackend
 from ..credentials import SimpleCredential
-from ..errors import PasswordDeleteError, ExceptionRaisedContext
+from ..errors import PasswordSetError,PasswordDeleteError,ExceptionRaisedContext
 
 
 with ExceptionRaisedContext() as missing_deps:
@@ -100,7 +100,12 @@ class WinVaultKeyring(KeyringBackend):
                           CredentialBlob=password,
                           Comment="Stored using python-keyring",
                           Persist=win32cred.CRED_PERSIST_ENTERPRISE)
-        win32cred.CredWrite(credential, 0)
+        try:
+            win32cred.CredWrite(credential, 0)
+        except pywintypes.error as e:
+            import sys
+            type, exc, tb = sys.exc_info()
+            raise PasswordSetError,("Max password length exceeded",type,exc),tb
 
     def delete_password(self, service, username):
         compound = self._compound_name(username, service)


keyring/backends/Windows.py
```diff
@@ -100,7 +100,12 @@ def _set_password(self, target, username, password):
                           CredentialBlob=password,
                           Comment="Stored using python-keyring",
                           Persist=win32cred.CRED_PERSIST_ENTERPRISE)
-        win32cred.CredWrite(credential, 0)
+        try:
+            win32cred.CredWrite(credential, 0)
+        except pywintypes.error as e:
+            import sys
+            type, exc, tb = sys.exc_info()
+            raise PasswordSetError,("Max password length exceeded",type,exc),tb
```
jaraco, line 108: How do you know "password length exceeded" is the cause? Couldn't other causes have triggered a `pywintypes.error`? I'd like to see some test here and just raise the original exception if it isn't indicating a password of too great a length.

keyring/backends/Windows.py
```diff
@@ -100,7 +100,12 @@ def _set_password(self, target, username, password):
                           CredentialBlob=password,
                           Comment="Stored using python-keyring",
                           Persist=win32cred.CRED_PERSIST_ENTERPRISE)
-        win32cred.CredWrite(credential, 0)
+        try:
+            win32cred.CredWrite(credential, 0)
+        except pywintypes.error as e:
+            import sys
```
jaraco, line 106: Please move this import to the header.

keyring/backends/Windows.py
jaraco, line 108: Probably this needs to use [six.reraise](https://pythonhosted.org/six/#six.reraise) for Python 3 compatibility.

jaraco (changes requested):
Thanks for working on this. You're on the right track, but the tests are failing due to style and syntax issues. I suggest fixing those but also, can you add a test? Ideally, there would be a test prior to the change that illustrates detection of the undesirable failure (fails) and the validates the fix (passing after the change is applied).

keyring/backends/Windows.py
cowlinator: Does keyring already have six as a dependency?  Or are you saying I should add it as one?

jaraco: Oh, interesting. I was assuming six was a dependency, but I see now it's not. I see that the py27compat module has some code borrowed from six.py, so probably best to repeat that pattern and copy [these lines](https://github.com/benjaminp/six/blob/6845a1b7670aeea9f5b19c3bf84aa6419f047a77/six.py#L684-L716) into py27compat.py.

jaraco:
Closing as stale, but feel free to update here and I can re-open if your interest is renewed.

Issue #375, opened 2019-01-14
Denying keyring popup on OSX raises platform-specific exception, making it hard to write cross-platform code


addibble:
I am trying to write an app using keyring that works on Linux with SecretService as well as OSX. SecretService raises KeyringLocked which inherits from keyring.errors.KeyringError when a resource is locked. On OSX, if you click "Deny" on the keyring dialog, keyring.backends._OS_X_API.Error is raised (which inherits directly from Exception). If keyring is installed on Linux, I can't catch this exception because I get this error:
AttributeError: module 'keyring.backends' has no attribute '_OS_X_API'
It would be better if when you click Deny on OSX, it raised KeyringLocked or at least something that inherits from KeyringException so I could catch this one class and the code would work on both platforms.

jaraco:
Great description. I agree - it should be possible and easy to catch such errors. I can imagine a couple of ways this might be accomplished:

- In _OS_X_API, raise a subclass of keyring.errors.KeyringError. I don't like this approach because I think of that module as being largely an independent module and shouldn't have a dependency on keyring.
- In OS_X.py, trap errors and replace them with a subclass of KeyringError. That seems reasonable to me, though it might be messy.

There may be other options too, though I'd probably start by drafting the latter solution and see what inspiration strikes.

@addibble - would you be interested in drafting a solution?

addibble:
Yes I'd be happy to write a fix. I looked into it a bit and I think the most correct fix would be to update _OS_X_API.py to detect this condition and raise a new platform-specific error class for this event (like _OS_X_API.KeychainDenied, so it can be differentiated from Error or NotFound) and have OS_X.py catch that and re-raise KeyringLocked, as well as catching _OS_X_API.Error in each OS_X.Keyring method and re-raise KeyringError so all API failures can be caught by the cross-platform exceptions.

If this seems acceptable please advise and I will submit a pull request.

jaraco:
Sounds great to me. And thanks for the report and in advance for any contribution you share.

PR #376, opened 2019-01-19
Fix for #375 add more specific error messages from _OS_X_API and re-raise more spe…

addibble:
…cific cross-platform exceptions from OS_X backend


addibble:
Should probably be merged with a major version bump IMO because the exceptions have changed

jaraco:
Mentioning #375 in text to get the link.

PR #379, opened 2019-03-05
add license file

jameslamb:
@jaraco I was messing around with this cool project called [raimon49/pip-licenses](https://github.com/raimon49/pip-licenses) tonight and found that this project shows up with license UNKNOWN when it is actually MIT.

I think this is because `pip-licenses` scans the license files bundled with packages you've installed in to `site-packages/` and you are not currently bundling your license file with your package on PyPi.

I hope you'll consider this PR to add the license to your source distribution.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+include LICENSE
+


jaraco:
Closing for the same reason as jaraco/keyrings.alt#34. If there is an issue not in pip-licenses, it may be best to file it with jaraco/skeleton, as these projects derive from there, and there are many other projects I maintain that would be subject to the same behavior.

PR #381, opened 2019-03-05
Fix #380

### LINKED ISSUE #380: keyring.backends.SecretService doesn't neatly close open sockets.
When I run a test_suite using setuptools test command and the nose.collector test_suite I get a couple of

    ResourceWarning: unclosed <socket.socket fd=4, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, raddr=/run/user/[...uid...]/bus>

I've tracked it down to the dbus socket being opened, but not closed before garbage collection.

CharString:
Turned `get_preferred_collection` into a context_manager.
This required no api change. I didn't take the effort to write a unittest that showed the issue.

### DIFF
--- a/keyring/backends/SecretService.py
+++ b/keyring/backends/SecretService.py
@@ -1,4 +1,5 @@
 import logging
+from contextlib import contextmanager
 
 from ..util import properties
 from ..backend import KeyringBackend
@@ -33,11 +34,13 @@ class Keyring(KeyringBackend):
         try:
             bus = secretstorage.dbus_init()
             list(secretstorage.get_all_collections(bus))
+            bus.sock.close()
         except exceptions.SecretStorageException as e:
             raise RuntimeError(
                 "Unable to initialize SecretService: %s" % e)
         return 5
 
+    @contextmanager
     def get_preferred_collection(self):
         """If self.preferred_collection contains a D-Bus path,
         the collection at that address is returned. Otherwise,
@@ -56,39 +59,42 @@ class Keyring(KeyringBackend):
             collection.unlock()
             if collection.is_locked():  # User dismissed the prompt
                 raise KeyringLocked("Failed to unlock the collection!")
-        return collection
+        try:
+            yield collection
+        finally:
+            collection.connection.sock.close()
 
     def get_password(self, service, username):
         """Get password of the username for the service
         """
-        collection = self.get_preferred_collection()
-        items = collection.search_items(
-            {"username": username, "service": service})
-        for item in items:
-            if hasattr(item, 'unlock'):
-                item.unlock()
-            if item.is_locked():  # User dismissed the prompt
-                raise KeyringLocked('Failed to unlock the item!')
-            return item.get_secret().decode('utf-8')
+        with self.get_preferred_collection() as collection:
+            items = collection.search_items(
+                {"username": username, "service": service})
+            for item in items:
+                if hasattr(item, 'unlock'):
+                    item.unlock()
+                if item.is_locked():  # User dismissed the prompt
+                    raise KeyringLocked('Failed to unlock the item!')
+                return item.get_secret().decode('utf-8')
 
     def set_password(self, service, username, password):
         """Set password for the username of the service
         """
-        collection = self.get_preferred_collection()
-        attributes = {
-            "application": self.appid,
-            "service": service,
-            "username": username
-        }
-        label = "Password for '%s' on '%s'" % (username, service)
-        collection.create_item(label, attributes, password, replace=True)
+        with self.get_preferred_collection() as collection:
+            attributes = {
+                "application": self.appid,
+                "service": service,
+                "username": username
+            }
+            label = "Password for '%s' on '%s'" % (username, service)
+            collection.create_item(label, attributes, password, replace=True)
 
     def delete_password(self, service, username):
         """Delete the stored password (only the first one)
         """
-        collection = self.get_preferred_collection()
-        items = collection.search_items(
-            {"username": username, "service": service})
-        for item in items:
-            return item.delete()
-        raise PasswordDeleteError("No such password!")
+        with self.get_preferred_collection() as collection:
+            items = collection.search_items(
+                {"username": username, "service": service})
+            for item in items:
+                return item.delete()
+            raise PasswordDeleteError("No such password!")
--- a/tox.ini
+++ b/tox.ini
@@ -9,6 +9,8 @@ commands =
 	pytest {posargs}
 usedevelop = True
 extras = testing
+passenv = 
+    DBUS_SESSION_BUS_ADDRESS
 
 [testenv:build-docs]
 extras =


keyring/backends/SecretService.py
```diff
@@ -56,39 +59,42 @@ def get_preferred_collection(self):
...
-        collection = self.get_preferred_collection()
-        items = collection.search_items(
-            {"username": username, "service": service})
-        for item in items:
-            if hasattr(item, 'unlock'):
-                item.unlock()
-            if item.is_locked():  # User dismissed the prompt
-                raise KeyringLocked('Failed to unlock the item!')
-            return item.get_secret().decode('utf-8')
+        with self.get_preferred_collection() as collection:
```
jaraco, line 70: Given how repetitive this change is and how it's creating a whole new layer of indentation for the entire function call, I'd rather see a decorator on each method that handles the context manager. I don't feel strongly enough about it that it needs to be part of this PR, but it's something I'd probably change after acceptance.

keyring/backends/SecretService.py
```diff
@@ -56,39 +59,42 @@ def get_preferred_collection(self):
             collection.unlock()
             if collection.is_locked():  # User dismissed the prompt
                 raise KeyringLocked("Failed to unlock the collection!")
-        return collection
+        try:
```
jaraco, line 62: This block should instead use `with contextlib.closing(collection.connection.sock)`

keyring/backends/SecretService.py
```diff
@@ -33,11 +34,13 @@ def priority(cls):
         try:
             bus = secretstorage.dbus_init()
             list(secretstorage.get_all_collections(bus))
+            bus.sock.close()
```
jaraco, line 37: I feel uneasy about this change. It's not obvious to me when `bus.sock` gets opened... or when it should be closed. Does it only need to be closed when iterating on `get_all_collections` succeeds? That's what this does.

jaraco (commented):
Overall, these changes seem reasonable. I'm unsure if both changes (there are two places in the code where a socket gets closed) are needed. An explanation for that would be helpful, as the bug report only reports one issue.

My main concern is the one raised in the parent ticket, that @mitya57 suggests these changes should be in the upstream library. That makes sense to me, especially since these changes seem to be reaching into the internal interfaces of the pertinent objects. Let's explore that first.

mitya57:
The socket is opened in [Jeepney’s `connect_and_authenticate` function][1].

It is wrong to close it in Keyring code, because Keyring still supports SecretStorage 2.x (under Python < 3.5) which does not use Jeepney at all, so this code will break there.

In my opinion the proper fix would be one of:

- Adding a destructor (`__del__` method) to Jeepney’s `DBusConnection` class which will close the socket.

- Turning SecretStorage’s `dbus_init` function into a context manager which will close the socket on exit, something like:
    ```python
    with secretstorage.dbus_init() as connection:
        collection = secretstorage.get_default_collection(connection)
        ...
    ```
    This will require changes in Keyring to use the context manager when it is available, and the traditional `dbus_init()` function for older versions of SecretStorage.

The first way will be easier to implement. @jaraco What will you prefer? @takluyver As Jeepney is concerned, I would like to know your opinion too.

[1]: https://gitlab.com/takluyver/jeepney/blob/master/jeepney/integrate/blocking.py#L71

takluyver:
Thanks, I've got plans to rework the integration layer when I get time, so this is useful to see. I'll think about what I can do. In the meantime, secretstorage is probably a better place to fix this.

mitya57:
I have added a [secretstorage.create_connection](https://secretstorage.readthedocs.io/en/latest/#secretstorage.create_connection) context manager today.

@jaraco Please let me know what you think about it. If it looks good to you then I will make a new release and then we will be able to update this merge request.

CharString:
@mitya57 Why the extra context manager? Why not make the `Collection` a context manager (much like the file-objects are)? Just adding `__enter__()`, ` __exit__()` and `close()` for when you have long-lived usage and manually want to close at some other time should do the trick.

All the current `get_` and `create_` will function a bit like the built-in `open`

mitya57:
I wanted to make it possible to re-use the same connection for operations with multiple connections.

Also I do not want to change current API: now functions like `get_collection_by_alias` and `create_collection` require `connection` as their first argument.

Removing that argument would be an API break. Making it possible to pass None doesn’t sound nice too.

jaraco:
@mitya57: Happy for Keyring to have alternate behaviors based on the available interfaces in the backends _especially_ if we're able to indicate one behavior as preferred (Jeepney/context manager) and one as deprecated/legacy with some sunset period.

mitya57:
We will be able to drop support for old SecretStorage versions if we drop support for Python 2 (and 3.4).

Are you going to do that anytime soon?

jaraco:
I've opened #383. It shouldn't be rushed, but the time has come to drop support for older versions.

CharString:
@mitya57 You don't have to remove the argument. Just give it a None default. And if is None, do dbus_init(). Just like builtin file-objects became context-managers without breaking existing code. You don't *have* to use the context-manager, but you can.

mitya57:
@CharString Consider the case of `create_collection()` function. I can make the `connection` argument optional and create it on the fly. But then I won’t be able to close it in the same function, because the returned `Collection` object will have a link to it. And without connection, you won’t be able to use that returned object.

I cannot close the connection in `Collection` destructor as well, as there may be `Item`s retrieved from that collection that will be using it.

Though now I realize that my approach suffers from this issue as well:
```python
>>> with secretstorage.create_connection() as connection:
...     collection = secretstorage.get_default_collection(connection)
... 
>>> list(collection.search_items({'application': 'myapp'}))
  File "<stdin>", line 1, in <module>
  File "/home/dmitry/upstream/secretstorage/secretstorage/collection.py", line 91, in search_items
    result, = self._collection.call('SearchItems', 'a{ss}', attributes)
  File "/home/dmitry/upstream/secretstorage/secretstorage/util.py", line 61, in call
    return self.send_and_get_reply(msg)
  File "/home/dmitry/upstream/secretstorage/secretstorage/util.py", line 47, in send_and_get_reply
    return self._connection.send_and_get_reply(msg)
  File "/home/dmitry/.local/lib/python3.7/site-packages/jeepney/integrate/blocking.py", line 45, in send_and_get_reply
    future = self.send_message(message)
  File "/home/dmitry/.local/lib/python3.7/site-packages/jeepney/integrate/blocking.py", line 26, in send_message
    self.sock.sendall(data)
OSError: [Errno 9] Bad file descriptor
```

I think I should keep that context manager as a temporary feature, and the issue should be fixed in Jeepney.

takluyver:
I can provide a more official way to close the connection in Jeepney, but I think it still leaves you with essentially the same issue for the secretstorage API.

Maybe the collection object itself should be a context manager, which closes the connection (if it created it) when it exits, so it's only usable inside the context.

mitya57:
So it looks like I have two choices:

- Completely change all SecretStorage API to make it not possible to have hanging Collection or Item instances when the context manager is exited. This means either moving Collection and Item methods to a single class, which will be the context manager, or making all methods explicitly require a `connection` argument.

- Make connections absolutely transparent for user, i.e. create a new connection for every single operation and then immediately close it. This will be more convenient for users but slower.

Is there some other way I missed?

In any case, having a more official way to close the connection won’t hurt.

takluyver:
Given what it does, I would imagine that secretstorage is rarely accessed many times in quick succession. As far as I know, secrets data doesn't change rapidly, so if code needs a secret inside a loop it makes sense to retrieve it once and cache it locally.

So possibly making a new connection for each operation and closing it again is a reasonable default. For 'power users', you could offer an optional `connection=` parameter to reuse an existing connection, and/or a context manager to keep a collection open for multiple operations.

CharString:
Hi,

Just now, over lunch, I hacked together https://github.com/CharString/secretstorage/commits/master
I tried to make easy things easy, and keep hard things possible. I think this is what @takluyver had in mind too.

So the one off use-case is like this:
 
    with secretstorage.Collection() as collection:
        for item in collection.get_all_items():
            print(item.get_label())  # Should this be the __str__ behaviour?

The "old" API works unchanged for "power users". With the added bonus of a 'close()' method that closes the socket without having to "reach in" to DBusConnection.sock... (Maybe it should have been named _sock, to convey it's "private" nature.)

To get this to work without changing Jeepney, I've implemented a temporary Wrapper for the DBusConnection that adds the `close()`, `__enter_()` and `__exit__()`. That of course should be done in Jeepney.

jaraco:
I appreciate the work on this PR and the associated ticket. I think this work is abandoned in favor of another tack, so I'm going to close it, but feel free to revive it or start a new one as appropriate.

mitya57:
Yeah, I have this on my TODO list but still don't have time for it :-(

PR #384, opened 2019-03-17
Drop support for EOL Python 2.7 - 3.4

### LINKED ISSUE #383: Drop support for Python 2.7
It is time. Let's make sure there aren't any critical bugs outstanding in the master and then cut a new release for Python 3.5+ or Python 3.6+.

hugovk:
Fixes #383.

Python 3.4 went EOL yesterday (2019-03-16).


.travis.yml
```diff
@@ -1,9 +1,7 @@
 dist: xenial
-sudo: false
```
jaraco, line 2: I prefer to run with `sudo: false`. Is that the default now? If so, I'd like to make that change in jaraco/skeleton.

keyring/backends/SecretService.py
```diff
@@ -80,7 +80,7 @@ def set_password(self, service, username, password):
             "service": service,
             "username": username
         }
-        label = "Password for '%s' on '%s'" % (username, service)
+        label = "Password for '{}' on '{}'".format(username, service)
```
jaraco, line 83: I probably would have deferred these compatible changes for another time, as it's more about style than compatibility, but happy to accept with them.

jaraco (changes requested):
All this needs now is a CHANGES.txt entry with a 1.x bump.

.travis.yml
hugovk: The `sudo: false` env is no longer available and the `sudo` keyword is fully deprecated:

* https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

Would you like me to remove this commit?

hugovk:
> All this needs now is a CHANGES.txt entry with a 1.x bump.

Updated!

.travis.yml
jaraco: Fine to leave it as submitted. Thanks for the reference.

jaraco (approved):

Issue #382, opened 2019-03-07
win32cred.CRED_PERSIST_ENTERPRISE and other keychain options


earonesty:
this option should be selectable by the user :

win32cred.CRED_PERSIST_ENTERPRISE

likewise (roaming or not roaming) should be available for other platforms.

1. i can implement this on windows.   not sure what the equivalent is on mac/linux:

2. I'm not sure what the interface should be even if i knew:

maybe something like this:

returns a dict with all supported options for that backend, or their defaults if the keychain has never been used

`res = keyring.get_chain_options("mykeychain")`

this likely will only work if the specified keychain has never been used before, raises a ValueError/NotSupported error if the platform doesn't support that option

`keyring.set_chain_options("mykeychain", roaming=False)`

jaraco:
To help me understand, why would a person want to change this value?

earonesty:
Suppose you have information that cannot leave the machine, like a machine private key in a multi-factor authentication scheme.   And then that factor magically migrates to another box....because windows.   We monkey patch to turn it off now.

jaraco:
Aah. That clarifies. I'm now leaning toward a solution based on environment variables, which could be set globally in the machine, per-user, per invocation, or set in the program.

mrx23dot:
Hello, 
How to actually use it?

```
keyring.set_chain_options("persist", 'local machine')
keyring.set_password('a', 'a', 'a')
```
didn't work. I couldn't find any documentation
Env.var also didn't work:
`set .persist="session"`

jaraco:
To set it on the backend itself, set the `.persist` attribute:

```
backend = keyring.core.get_keyring()
backend.persist = 'session'
# or
backend.persist = win32cred.CRED_PERSIST_SESSION
```

Because keyring has a [general feature](https://github.com/jaraco/keyring#altering-keyring-behavior) that sets values on the backend, you can use that too. Set the environment variable `KEYRING_PROPERTY_PERSIST='session'`.

Let me know how that works for you. And if some documentation would have helped somewhere, please consider contributing that to help the next person.

mrx23dot:
Works great, thanks!

PR #385, opened 2019-03-21
minimum needed to control windows roaming

earonesty:
This is a minimal solution.   It requires the user to know about the issue and choose to address it in a platform specific if block.

https://github.com/jaraco/keyring/issues/382

Until there are many such options, having an options framework may be overkill.

### DIFF
--- a/keyring/backends/Windows.py
+++ b/keyring/backends/Windows.py
@@ -83,23 +83,23 @@ class WinVaultKeyring(KeyringBackend):
             raise
         return res
 
-    def set_password(self, service, username, password):
+    def set_password(self, service, username, password, persist_type=win32cred.CRED_PERSIST_ENTERPRISE):
         existing_pw = self._get_password(service)
         if existing_pw:
             # resave the existing password using a compound target
             existing_username = existing_pw['UserName']
             target = self._compound_name(existing_username, service)
             self._set_password(target, existing_username,
-                               existing_pw['CredentialBlob'].decode('utf-16'))
-        self._set_password(service, username, text_type(password))
+                               existing_pw['CredentialBlob'].decode('utf-16'), persist_type)
+        self._set_password(service, username, text_type(password), persist_type)
 
-    def _set_password(self, target, username, password):
+    def _set_password(self, target, username, password, persist_type):
         credential = dict(Type=win32cred.CRED_TYPE_GENERIC,
                           TargetName=target,
                           UserName=username,
                           CredentialBlob=password,
                           Comment="Stored using python-keyring",
-                          Persist=win32cred.CRED_PERSIST_ENTERPRISE)
+                          Persist=persist_type)
         win32cred.CredWrite(credential, 0)
 
     def delete_password(self, service, username):
--- a/keyring/core.py
+++ b/keyring/core.py
@@ -57,10 +57,10 @@ def get_password(service_name, username):
     return _keyring_backend.get_password(service_name, username)
 
 
-def set_password(service_name, username, password):
+def set_password(service_name, username, password, **backend_opts):
     """Set password for the user in the specified service.
     """
-    _keyring_backend.set_password(service_name, username, password)
+    _keyring_backend.set_password(service_name, username, password, **backend_opts)
 
 
 def delete_password(service_name, username):


jaraco:
Thanks for working on a patch.

I wonder if there's a better way to solicit these options than to accept parameters to the `set_password` function, which really shouldn't accept parameters.

I can think of a couple of other ways that would be preferable: either with a class attribute that can be overridden (perhaps by a subclass) or with an environment variable or with a configuration setting. Currently, the configuration file isn't used to set options, but it seems look a good place for something that would affect a user or environment (and not just a single program).

earonesty:
This is something that needs to be set on a per-program basis, not per-user.   The user already has the ability to turn off roaming of his profile at the O/S level.   

I did originally imagine a "set_options(keychain_name, roaming=False)" top level function that take kwargs and raises an exception if the option isn't supported.

Just as easy to implement.

jaraco (changes requested):
Here's what I'd like to see:

Abstract away the 'win32cred' implementation details and instead expose a name that expresses to the user the effect they're seeking, something like "persist" or "scope", which could be "transient", "local", or "roaming" (mapping to the three possible values for Persist).

Then, expose this scope through an environment variable, something like `scope = lookup[os.environ.get('KEYRING_SCOPE', 'roaming')]` where `lookup = dict(roaming=win32cred.CRED_PERSIST_ENTERPRISE, ...)`. This logic could happen in a nice encapsulated private function with a docstring explaining the purpose.

Also, this change definitely deserves an update to the changelog explaining what it does and how it's used and why one would use it.

earonesty:
The terminology abstraction idea is solid.  However, it's possible a single application can have legitimate use for roaming, local and transient keychain access.    An env var won't give you the granularity that one might need.    Sees like there needs to be some sort of top level functional access to keychain options.   You could get away with a variable env var  `KEYRING_SCOPE_<keychainname>.`   

But it seems better to have `set_options(keychain_name, scope="roaming")` ... which shoves these into a dict of some kind that backends can access.   Still not sure if it's better to fail if the option isn't supported or quietly ignore the option and continue with existing behavior.  Of course since this isn't "stored" in the keychain in any way, it doesn't help with command line use.    So maybe the env var is ok, and a python application can just change scope before and after use?

jaraco:
> it's possible a single application can have legitimate use for [different options in the same session].

Do you encounter this use-case in practice? Can you illustrate such a use-case? Couldn't still the application do something like:

```
from unittest.mock import patch

patch(os.environ, KEYRING_SCOPE='local')(keyring.set_password)(system, user, password)
```

or perhaps more readable:

```
with patch(os.environ, KEYRING_SCOPE='local'):
    keyring.set_password(system, user, password)
```

It's not super-elegant, nor thread-safe, but it provides access for setting a local password for the duration of a single call.

*edit*

> So maybe the env var is ok, and a python application can just change scope before and after use?

Aah right, that's what you were suggesting. Yes, I think that's suitable.

earonesty:
I'd prefer it if the env var was used for the default, but the caller could still override it.   Thread safety is an issue here, we have GUI threads setting roaming passwords, while machine-specific (non roaming) device private keys are being accessed in other threads.   Would an env var, and the caller can pass an optional env={...} dictionary keyword arg that contains overrides, or do you really, really not want to add an optional arg?

jaraco:
Sorry to have let this fall off my radar without answering your question.

> Would an env var, and the caller can pass an optional env={...} dictionary keyword arg that contains overrides, or do you really, really not want to add an optional arg?

The danger with allowing parameters to be passed is that there's not a designed/sanctioned mechanism for a backend to solicit parameters from the caller or command-line. I worry about `backend_opts` because it starts to establish that design, but only for the `set` operation. It doesn't concern itself with `get` or `delete`, which may also need parameters. Moreover, other keyrings aren't soliciting `**backend_opts`, so now the ABC suggests the backends might solicit arbitrary keyword arguments, but only one does. And then we have to ask ourselves if these backend options should be exposed through the CLI as well... and what happens when a particular option is supplied that's not relevant to a backend.... or what happens when another backend like the ChainerBackend is involved.

We've had to deal with similar customization issues in the macOS backend, which might direct the user to one of a couple of [stores in the Keychain](https://github.com/jaraco/keyring/blob/979f4dd6078b7ee2db3d5ac71dd3e5bfad62331e/keyring/backends/OS_X.py#L19) (generic or internet, I think).

My motivation for suggesting the environment variable was to avoid affecting the command-line and function calls in the API. In retrospect, the thread safety concerns and awkwardness of that approach really isn't suitable.

It's for these reasons that I'd prefer if the solution were to follow the same model of the macOS backend and have customizations be set on the class (or instance) and not as parameters to method(s). Such an approach should be threadsafe (as separate backends can be constructed/customized in the same thread). Something like what I've drafted in cc71900. What do you think of that approach?

jaraco:
In addition to the change I've proposed, I'm thinking keyring should solicit `KEYRING_PROPERTY_*`, which it will set on any keyring. So for this design, you could set KEYRING_PROPERTY_PERSIST=LOCAL_MACHINE to get the local machine behavior by default.

jaraco:
In #401 and #407 (releasing as 19.3), I've implemented two features that provide interfaces for supporting scopes for Windows. It's now possible through an env var or programmatically to set a `.persist` property on the keyring to alter the persistence scope. If you think it's worthwhile for the CLI to also allow for setting keyring properties on a per-invocation basis, I'd be willing to accept a change in that regard (should behave similar to environment variable, setting arbitrary properties on a backend).

Issue #217, opened 2016-03-14
Restore ctypes implementation for OS X backend


jaraco:
There have been several issues relating to the implementation of the OS X Keyring delegating to the security command:
- #215 Tests fail intermittently on OS X
- #145 Password exposure through process listing on OS X
- #144 Keyring access control on OS X

Additionally, there are other unreported issues, such as the workaround for #145 only working for some passwords but not others.

As a result, I've started work on a pure ctypes implementation in the [osx-api branch](https://github.com/jaraco/keyring/commits/osx-api). 

Subsequently, there are several issues that should be revisited before reverting back to a CFFI or ctypes interface:
- This current implementation drops support added in #73, which would have to be restored (or explicitly dropped).
- Confirm reasonable virtualenv behavior, respecting #13 OSX install broken when using virtualenvs.
- Confirm that there's no regression w.r.t. #11 Should not use login.keychain on OSX.

My plan is to address these issues and run the branch version locally for some time to understand the impacts of these changes. I welcome others to do the same or provide feedback.

copumpkin:
I was about to file a bug on exactly this, since it seemed pretty bad that the permission was coming from `security` (i.e., #144). A thousand :+1: from me.

jaraco:
> This current implementation drops support added in #73, which would have to be restored (or explicitly dropped).

Given the non-response to the release of keyring 2.1 in #10 and #73, I'm not even sure that feature was ever used. To complicate matters, the API is not congruent, [requiring many more parameters](https://github.com/jaraco/keyring/blob/dc793d1/keyring/backends/_OS_X_API.py#L47-L65) than the generic password counterpart. I'm going to add a warning to the released versions to see if there are any users relying on this functionality.

jaraco:
> Confirm reasonable virtualenv behavior, respecting #13 OSX install broken when using virtualenvs.

I've confirmed that keyring does work naturally in Python 3 venv environments, and in `virtualenv` environments, the user is prompted to Allow or Deny or Always Allow "python" to access the keyring. In other words, this change fixes #144 and regresses w.r.t. #13. Overall, I think the change is beneficial, especially in light of copumpkin's endorsement. It's clear that a backward-incompatible release will be necessary, however.

jaraco:
> Confirm that there's no regression w.r.t. #11 Should not use login.keychain on OSX.

This is now done, with the default keyring now being honored by default.

jaraco:
[...]

PR #394, opened 2019-06-18
Made username optional with get commands

joshua-johnson-cfa:
On OS_X and SecretService, I've implemented some logic to make the username parameter optional in get_password/get_credential. Also the get password function on OS_X will check the service name to see if it should use find_generic or find_internet.

### DIFF
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -127,9 +127,10 @@ class KeyringBackend(metaclass=KeyringBackendMeta):
         returned username.
         """
         # The default implementation requires a username here.
-        cred = self.get_password(service, username)
-        if cred['password'] is not None:
-            return credentials.SimpleCredential(cred['username'], cred['password'])
+        if username is not None:
+            password = self.get_password(service, username)
+            if password is not None:
+                return credentials.SimpleCredential(username, password)
         return None
 
 
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -127,10 +127,9 @@ class KeyringBackend(metaclass=KeyringBackendMeta):
         returned username.
         """
         # The default implementation requires a username here.
-        if username is not None:
-            password = self.get_password(service, username)
-            if password is not None:
-                return credentials.SimpleCredential(username, password)
+        cred = self.get_password(service, username)
+        if cred['password'] is not None:
+            return credentials.SimpleCredential(cred['username'], cred['password'])
         return None
 
 
--- a/keyring/backends/OS_X.py
+++ b/keyring/backends/OS_X.py
@@ -45,14 +45,11 @@ class Keyring(KeyringBackend):
 
     def get_password(self, service, username):
         creds={}
+        type = get_service_type(service)
         if (not username):
-            prompt = "find-generic-password"
-            
-            if (service.find(".com")>=0):
-                prompt = "find-internet-password"
             
             output = execute(
-                    'security 2>&1 '+prompt+' -g -s '+service,
+                    'security 2>&1 '+type+' -g -s '+service,
                     lambda x: "",
                     lambda x: "",
                     ignore_exit_codes=True
@@ -65,7 +62,7 @@ class Keyring(KeyringBackend):
 
         else:
             try:
-                if (service.find(".com")>=0):
+                if (type == "find-internet-password"):
                     creds['password'] = api.find_internet_password(self.keychain, service, username)
                 else:
                     creds['password'] = api.find_generic_password(self.keychain, service, username)
@@ -159,3 +156,9 @@ def execute(cmds, stdout_cb, stderr_cb, env=None, ignore_exit_codes=False, print
 def find_key(fn, out):
     match = fn(out)
     return match and match.group(1)
+
+def get_service_type(service):
+    """Helper function to determin what type of password to look up"""
+    if ("https://" in service or "http://" in service):
+        return "find-internet-password"
+    return "find-generic-password"
--- a/keyring/backends/OS_X.py
+++ b/keyring/backends/OS_X.py
@@ -46,21 +46,29 @@ class Keyring(KeyringBackend):
     def get_password(self, service, username):
         creds={}
         if (not username):
+            prompt = "find-generic-password"
+            
+            if (service.find(".com")):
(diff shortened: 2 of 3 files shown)


jaraco:
Sorry about the issues with the linter. I've [reported that issue with the linter](https://github.com/tholo/pytest-flake8/issues/42#issuecomment-504549459) so that will I hope be addressed in the future.

In the meantime, I'll review what I can of the code.

keyring/backend.py
```diff
@@ -127,10 +127,9 @@ def get_credential(self, service, username):
         returned username.
         """
         # The default implementation requires a username here.
-        if username is not None:
-            password = self.get_password(service, username)
-            if password is not None:
-                return credentials.SimpleCredential(username, password)
+        cred = self.get_password(service, username)
```
jaraco: I don't think this code will work across all backends. The interface for `.get_password` is to return a text string. The whole point of `get_credential` was to allow backends to retrieve a credential for a service without supplying a username. See the WinVaultKeyring for a non-fallback implementation of `get_credential`. Perhaps that's the method you wish to override in the macOS and Linux backends.

keyring/backends/OS_X.py
```diff
@@ -41,17 +44,40 @@ def set_password(self, service, username, password):
...
-        except api.Error as e:
-            raise KeyringError("Can't get password from keychain: " "{}".format(e))
+            find_passwd = re.compile('password: "([^"]+)"').search
+            find_user = re.compile('"acct"<blob>="([^"]+)"').search
+            creds['username'] = find_key(find_user, output)
+            creds['password'] = find_key(find_passwd, output)
+
+        else:
+            try:
+                if (service.find(".com")>=0):
```
jaraco: I'm pretty sure `.com` isn't a suitable indicator of whether a service represents an internet address.

keyring/backends/OS_X.py
```diff
@@ -41,17 +44,40 @@ def set_password(self, service, username, password):
             raise PasswordSetError("Can't store password on keychain: " "{}".format(e))
 
     def get_password(self, service, username):
-        if username is None:
-            username = ''
+        creds={}
+        if (not username):
+            prompt = "find-generic-password"
+            
+            if (service.find(".com")>=0):
```
jaraco: This logic looks like a copy/paste of line 68 below. Probably this functionality should be its own function or method with a docstring explaining the purpose.

keyring/backends/OS_X.py
```diff
@@ -41,17 +44,40 @@ def set_password(self, service, username, password):
...
+                prompt = "find-internet-password"
+            
+            output = execute(
```
jaraco, line 51: This backend seeks not to use the `security` executable to perform its duties, but instead perform its duties with the C APIs... and I definitely want to avoid mixing the two techniques. Can you accomplish the same thing with the APIs (either as implemented or by extending the ctypes adapters)?

jaraco (changes requested):
My comments above describe the changes I'm seeking.  If you can implement this without invoking subprocesses and without altering the base class or `get_password` interfaces, I think this could be an acceptable change. Thanks for working on it. Let me know if I can provide more guidance or clarity.

keyring/backend.py
joshua-johnson-cfa: I agree, I was going about this the wrong way. Restored this back to original logic.

keyring/backends/OS_X.py
joshua-johnson-cfa: I agree, that was a highly insufficient solution. I'll instead check for https or http

keyring/backends/OS_X.py
joshua-johnson-cfa: This is where I'm struggling. I've been scouring Apple's developer documentation for keychain, but of course its all in Objective-C so its only helpful to a certain point. I've been trying two separate approaches, so far neither have been successful.

Approach 1:
Currently you're using `SecKeychainFindGenericPassword` or `SecKeychainFindInternetPassword`. to retrieve the password. If an empty string is passed into the function for the username, the function still returns the password for the specified service, but not the corresponding username. Seeing that the last paramater of this function is an "itemRef" I tried passing this into `SecKeychainItemCopyAttributesAndData` and playing around with this function but the best I was able to do was return the password again.

Approach 2:
I noticed from Apple's documentation that `SecKeychainFindGenericPassword` is a legacy function and that they now recommend the use of `SecItemCopyMatching` which appears to be able to return a reference to all items associated with a keychain entry so in theory this seemed like the most logical approach, however I have thus far been unable to implement this successfully. It starts to get super messy trying when trying to cast things to the expected CF types.

joshua-johnson-cfa: Just out of curiosity what advantages do you find in having it implemented using the C APIs as opposed to executing the security shell commands?

jaraco: > what advantages do you find in having it implement using the C APIs?

See the description for #217, which details the motivations.

> I've been trying two separate approaches.

Thanks for detailing the attempts. I'll try to set some time aside in the future to explore the possibilities. The second approach sounds more stable long-term.

jaraco (commented):
Just following up on the conversation above.

keyring/backends/OS_X.py
jaraco: > `SecKeychainFindGenericPassword` is a legacy function

I confirmed that `SecKeychainFindGenericPassword` is indeed listed under [Legacy Password Storage](https://developer.apple.com/documentation/security/keychain_services/keychain_items?language=objc#1659107), referencing [Keychain Item Search](https://developer.apple.com/documentation/security/keychain_services/keychain_items?language=objc#1662299) (`SecItemCopyMatching`) as the replacement.

jaraco: I've created #400 to explore Approach 2. I still think Approach 1 has some merit, but I don't know when I'll find the time to flesh it out. My instinct, given the Legacy status, is to focus on Approach 2.

jaraco:
Since this PR is fraught with challenges, I'm going to close it, acknowledging that we may want to revisit the proposal once a mechanism for retrieving the username can be achieved.

PR #395, opened 2019-07-23
Add support for get_credentials() with the SecretService keyring

morguldir:


jaraco:
In general, this looks good to me. I suspect that `test_credential` is already covering the changes. All that's needed is an entry in the CHANGES.txt. Would you add a new entry with an 0.1 bump describing the change?

mitya57 (approved):

PR #404, opened 2019-10-17
Don't import all keyrings unless necessary

jyn514:
If a keyring is set in configuration or environment variables,
use that one and don't load all available keyrings.

Partially addresses https://github.com/jaraco/keyring/issues/403.

### DIFF
--- a/keyring/core.py
+++ b/keyring/core.py
@@ -87,13 +87,12 @@ def init_backend(limit=None):
     # save the limit for the chainer to honor
     backend._limit = limit
 
-    # get all keyrings passing the limit filter
-    keyrings = filter(limit, backend.get_all_keyring())
-
     set_keyring(
         load_env()
         or load_config()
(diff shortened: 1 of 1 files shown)


jyn514:
@mitya57 this has been sitting around for a while, are there any changes I should make?

mitya57:
Sorry for the delay. Thinking more about this, `filter()` is lazy in Python 3, so it does not actually do any iteration when you assign its result to a variable.

Maybe you are still using Python 2? We no longer support it…

jyn514:
Filter is lazy, but `get_all_keyrings()` is eager and is called before it is passed to filter. I am using Python3, I can show you the time speedups of you want.

jyn514:
If you want to keep the existing code structure, maybe `get_all_keyrings` could be turned into a generator so it is evaluated lazily.

mitya57:
That makes sense. Thank you! Maybe in future we should make `get_all_keyring` lazier too (and add the missing `s` to its name).

Issue #415, opened 2019-12-08
Keyring does not work with PyInstaller


toyg:
All details are in the sister bug at https://github.com/pyinstaller/pyinstaller/issues/4569

Filing here as well because I think the hook they shipped was provided by you...? The hook is executed but somehow the `keyring.backends` module still does not appear in metadata.

jaraco:
This is a duplicate of #324.

Issue #324, opened 2018-05-09
Keyrings unavailable on pyinstaller


c-rei:
This error appears when I updated keyring from version 11.0.0 to version 12.2.0 and compiled my source with the pyinstaller. The source file works fine, However on running the binary I get `RuntimeError: No recommended backend was available. Install the keyrings.alt package if you want to use the non-recommended backends. See README.rst for details.`

I have installed keyrings.alt, but I don't known how to use it, and how to modify my source for keyring.

Thank for any help.

jaraco:
You'll notice in the changelog that [starting with version 12 of keyring](https://keyring.readthedocs.io/en/latest/history.html), it now relies on entry points to discover keyring backends. My suspicion is that something about pyinstaller breaks the metadata, rendering the backends invisible.

I think you have three options.

1. Stick with `keyring<12`.
2. In your application, explicitly import the keyring backend modules and re-initialize the keyring. Perhaps keyring could supply this functionality as a call for pyinstaller-based applications.
3. Verify my suspicion and then ask pyinstaller to support discovery of entry points in their environment.

c-rei:
Thank you for your help.
I have investigated the keyring backend with get_keyring function, and that returned ~Windos.WinVaultKering while I on my source, but ~fail.Keyring while I on a binary. So I'v tried to set the "keyring.backends.Windows.WinVaultKeyring" with set_keyring function, but got an error "AttributeError: module 'keyring.backend' has no attribute 'Windows'".
I want to choice the option 2 that you advised, but I don't known how to re-initialize the keyring. Could you give me some idea?

jaraco:
I think you want to `import keyring.backends.Windows` in order for `keyring.backends.Windows.WinVaultKeyring` to be viable. Does that help?

c-rei:
It works now as you advised.
Thank you very much.

genotrance:
I just spent several hours struggling with PyInstaller and keyring. I think it is worthwhile for keyring to provide functionality to pick the correct backend by default once frozen like you suggested rather than have to hard code it.

jaraco:
I'm re-opening this issue, as it remains an issue (as reported in the duplicate issues). In #416, here's what I recommend:

[...]

PR #416, opened 2019-12-10
Fix work with PyInstaller

msk849:
Fix for #415 issue

### DIFF
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -196,14 +196,36 @@ def _load_plugins():
 
     `initialize_func` is optional, but will be invoked if callable.
     """
-    try: entry_points = metadata.entry_points()['keyring.backends']
+    try:
+        entry_points = metadata.entry_points()['keyring.backends']
     except Exception:
         entry_points = (
-            metadata.EntryPoint(name='KWallet',value='keyring.backends.kwallet',group='keyring.backends'),
-            metadata.EntryPoint(name='SecretService',value='keyring.backends.SecretService',group='keyring.backends'),
-            metadata.EntryPoint(name='Windows',value='keyring.backends.Windows',group='keyring.backends'),
-            metadata.EntryPoint(name='chainer',value='keyring.backends.chainer',group='keyring.backends'),
-            metadata.EntryPoint(name='macOS',value='keyring.backends.OS_X',group='keyring.backends'))
+            metadata.EntryPoint(
+                name='KWallet',
+                value='keyring.backends.kwallet',
+                group='keyring.backends'
+            ),
+            metadata.EntryPoint(
+                name='SecretService',
+                value='keyring.backends.SecretService',
+                group='keyring.backends'
+            ),
+            metadata.EntryPoint(
+                name='Windows',
+                value='keyring.backends.Windows',
+                group='keyring.backends'
+            ),
+            metadata.EntryPoint(
+                name='chainer',
+                value='keyring.backends.chainer',
+                group='keyring.backends'
+            ),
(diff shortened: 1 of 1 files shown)


keyring/backend.py
```diff
@@ -196,7 +196,14 @@ def _load_plugins():
 
     `initialize_func` is optional, but will be invoked if callable.
     """
-    entry_points = metadata.entry_points()['keyring.backends']
+    try: entry_points = metadata.entry_points()['keyring.backends']
```
om26er: I think it would make sense to follow rest of the project and do try/catch using the "standard" syntax. Please change the code to
```
try:
    entry_points = metadata.entry_points()['keyring.backends']
```

om26er: Even better, there are suggestions on the CI failure page about what needs to be changed for the CI to pass https://ci.appveyor.com/project/jaraco/keyring/builds/29440604/job/r9f96fw8fh5a2f88 -- Kindly change the code as suggested. (Note: this is not a review from project maintainers).

keyring/backend.py
```diff
@@ -196,7 +196,34 @@ def _load_plugins():
 
     `initialize_func` is optional, but will be invoked if callable.
     """
-    entry_points = metadata.entry_points()['keyring.backends']
+    try:
+        entry_points = metadata.entry_points()['keyring.backends']
+    except Exception:
```
mitya57: Is it possible to catch a less generic exception here?

msk849: Yes. I will fix it

keyring/backend.py
msk849: Thanks. Now I have got CI pass.

mitya57 (approved):

om26er:
Since PyInstaller provides runtime information, I think it would make sense to explicitly check if the app is running from pyinstaller instead of the exception handling that is proposed. So something like:
```
import sys
if getattr(sys, 'frozen', False):
    pysintaller entry point fix goes here...
else:
    entry_points = metadata.entry_points()['keyring.backends']
```
more reading https://pythonhosted.org/PyInstaller/runtime-information.html#run-time-information

mitya57:
I like the current implementation more. It will work not only with PyInstaller but also with other freezers. Or when one uses python-keyring from git without installing it.

Let's see what @jaraco thinks.

jaraco:
I'm disinclined to accept this change because it is tackling the problem at the wrong point of abstraction. Although `keyring` is affected by the fact that pyinstaller loses the metadata for a package, hard-coding the entry-points into each and every application that supplies entry points can't be the solution. If I were to accept this change, that would also imply that other projects like `keyrings.alt` and `pmxbot` (with all its plugins) and others should follow suit. This change fails to support keyrings other than the backends supplied in this package and adds maintenance burden when backends are modified.

Instead, what needs to happen is to follow my advice in #324 - pyinstaller should supply the metadata. Now that keyring is using [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html), it should be able to follow the [instructions on extending the search algorithm](https://docs.python.org/3/library/importlib.metadata.html#extending-the-search-algorithm) to improve the discoverability of _all_ metadata (and not just entrypoints) for the packages it is assembling.

If hard-coding the entry points (and not implementing the feature directly in pyinstaller) is the recommended solution, then I'd like to see this hard-coding done outside of keyring itself, perhaps in a package like pyinstaller-all-entry-points, which itself could implement the `importlib.metadata` protocol to discover the hard-coded entry points, and would keyring to retain its simpler approach.

Issue #411, opened 2019-12-02
Import module and string formatting error


paulsbond:
**Describe the bug**
When using pip I see the a keyring error before anything happens, e.g.

```bash
$ python3 -m pip list | grep keyring
Keyring is skipped due to an exception: not all arguments converted during string formatting
keyring               19.3.0              
keyrings.alt          3.0
```

The error disappears if I remove the keyring package.
Using keyring gives more details of the error:

```bash
$ keyring --list-backends
Traceback (most recent call last):
  File "/home/paul/.local/lib/python3.6/site-packages/keyring/backend.py", line 203, in _load_plugins
    init_func = ep.load()
  File "/home/paul/.local/lib/python3.6/site-packages/importlib_metadata/__init__.py", line 92, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/usr/lib/python3/dist-packages/keyrings/alt/Windows.py", line 9, in <module>
    from . import file_base
  File "/usr/lib/python3/dist-packages/keyrings/alt/file_base.py", line 13, in <module>
    from keyring.util.escape import escape as escape_for_ini
ModuleNotFoundError: No module named 'keyring.util.escape'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/paul/.local/bin/keyring", line 5, in <module>
    from keyring.cli import main
  File "/home/paul/.local/lib/python3.6/site-packages/keyring/__init__.py", line 1, in <module>
    from .core import (
  File "/home/paul/.local/lib/python3.6/site-packages/keyring/core.py", line 192, in <module>
    init_backend()
  File "/home/paul/.local/lib/python3.6/site-packages/keyring/core.py", line 96, in init_backend
    filter(limit, backend.get_all_keyring()),
  File "/home/paul/.local/lib/python3.6/site-packages/keyring/util/__init__.py", line 22, in wrapper
[...]

PR #417, opened 2019-12-15
Fix log message formatting during exception handling

abn:
This fixes the log message by ensuring we only log the `EntryPoint` name. The root issue can be reproduced independently as shown below.

```sh
$ python -c 'from importlib.metadata import EntryPoint; "%s" % EntryPoint(name="name", value="value", group="group")'
TypeError: not all arguments converted during string formatting
```

Related to: #411 pypa/twine#552 python-poetry/poetry#1719

The above issues resolved the exception root causes but never fixed the second exception thrown from the exception handling block.

### DIFF
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -204,7 +204,7 @@ def _load_plugins():
             if callable(init_func):
                 init_func()
         except Exception:
-            log.exception("Error initializing plugin %s." % ep)
+            log.exception("Error initializing plugin %s." % ep.name)
 
 
 @util.once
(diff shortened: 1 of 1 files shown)


jaraco:
Thanks for this. This undesirable behavior was an unfortunate consequence of the fact that `EntryPoint` is iterable (a namedtuple), so `%s` string formatting tries to iterate it, rather than invoking `repr` on the object. I'll want to decide if rendering the entire entry point was the intended behavior and should be retained or if rendering just the name is sufficient.

jaraco:
Seeing as that code didn't change with either entrypoints replacement (a9fe22750359c2ca337c83d838e3654ef1fa38ea, 0028d732d78345f5f4818f5715f6054581bbf4b1), I think rendering the whole entrypoint is what's intended here.

PR #419, opened 2020-01-04
Close the secretstorage connections after using them

### LINKED ISSUE #380: keyring.backends.SecretService doesn't neatly close open sockets.

mitya57:
This fixes #380.

This relies on the `DBusConnection.close()` method that was added in yesterday's jeepney 0.4.2 release.

I am going to release a new version of SecretStorage that will depend on jeepney 0.4.2, and then bump the dependency on SecretStorage here.

### DIFF
--- a/keyring/backends/SecretService.py
+++ b/keyring/backends/SecretService.py
@@ -1,3 +1,4 @@
+from contextlib import closing
 import logging
 
 from ..util import properties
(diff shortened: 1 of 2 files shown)


techalchemy:
Hey @mitya57 I was the one who filed the initial bug over on jeepney, thanks for following this up with a PR!

This seems to work perfectly actually, ignore my prior comments! Didn't have the latest version of jeepney

mitya57:
@techalchemy Thanks for the feedback!
The idea is that the cleanup happens not in `get_preferred_connection` itself, but in methods that call it.

I'm glad it works fine for you now, however I don't know how could it fix *anything* if you had older jeepney installed.

Anyway, I am going to release new SecretStorage version and then remove Draft status from this PR in a few days.

PR #422, opened 2020-02-07
KeyError: 'keyring.backends'

yanone:
Not testing for 'keyring.backends' threw a KeyError in my setup. 
Not sure whether the key needs to exist, but testing for it works.

### DIFF
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -198,7 +198,7 @@ def _load_plugins():
     """
     metadata_entry_points = metadata.entry_points()
     if 'keyring.backends' in metadata_entry_points:
-        entry_points = metadata.entry_points()['keyring.backends']
+        entry_points = metadata_entry_points['keyring.backends']
         for ep in entry_points:
             try:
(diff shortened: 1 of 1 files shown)


jaraco:
Thanks for the contrib. I'd like for the problem to be better defined - in particular, under what conditions would 'keyring.backends' not be defined, have these conditions been supported in the past, and why should these conditions be supported going forward? If we can answer those concerns, then the next step is probably to either protect the code with a test or comment so that it doesn't get converted back to the current behavior.

Also, as a bit of a nitpick, I'd re-write the patch to avoid the extra branch and indentation, and instead use something like:

```
entry_points = metadata.entry_points().get('keyring.backends', ())
```

jaraco:
Happy to consider this change, but please, let's start with a problem description (in an issue) that describes the problem, under which conditions it occurs, and what justifies this change in behavior. Once we have agreement on the why, let's work on adding a test for this behavior. Thanks.

Issue #372, opened 2019-01-11
default backend is not deterministic


syrkuit:
When no backends are available, keyring.backends.chainer.ChainerBackend & keyring.backends.fail.Keyring end up having the same priority of 0
Because keyring.backend.get_all_keyring() is not deterministic, keyring.init_backend() isn't either as shown below.
As a result, behaviour is unpredictable, including raising a vague RuntimeError() exception:

```
$ repeat 10 python3 -c 'import keyring
print(keyring.backend.get_all_keyring()) 
try:                                                 
  print(keyring.get_password("not", "deterministic"))
except Exception as e:
  print(str(e))'
[<keyring.backends.chainer.ChainerBackend object at 0x7f914dcd85c0>, <keyring.backends.fail.Keyring object at 0x7f914ce7ad30>]
None
[<keyring.backends.fail.Keyring object at 0x7f8ed963d5f8>, <keyring.backends.chainer.ChainerBackend object at 0x7f8ed87dfc88>]
No recommended backend was available. Install the keyrings.alt package if you want to use the non-recommended backends. See README.rst for details.
[....]
```

syrkuit:
btw- this is Python 3.7.2, with debian python3-keyring 17.1.1-1

jaraco:
Indeed, that's undesirable. Thanks for the report.

jaraco:
One way to address this would be to set the priority lower than 0 on the chainer, though that doesn't feel like a robust approach.

syrkuit:
Yeah, that seemed like an easy way out, but not particularly clean which is why I didn't bother with a PR :-)

remd:
I am experiencing different behavior. In a shell for-loop I always get `keyring.backends.fail.Keyring` and in a Python REPL I always get `keyring.backends.chainer.ChainerBackend`. Perhaps this is not surprising if the behavior is not deterministic. 

I am using Python 3.6.7 and keyring 18.0.0 from pip. 

In repeat loop:
```
$ for i in {1..10}; do python -c 'import keyring
print(keyring.backend.get_all_keyring())
try:
  print(keyring.get_password("not", "deterministic"))
except Exception as e:
  print(str(e))'; done
[<keyring.backends.fail.Keyring object at 0x7f76466ed940>, <keyring.backends.chainer.ChainerBackend object at 0x7f76432b97f0>]
[...]

PR #424, opened 2020-02-21
without backends, always use fail backend, not empty chainer

ddstreet:
This lowers chainer's priority down to -10 if it has no backends, since it shouldn't be used - its behavior when empty is the same as the null backend.

This also introduces a new NoKeyringError class, which inherits from both KeyringError as well as RuntimeError, so it can be used in the fail backend, and keep backwards compatibility (as the fail backend previously raised RuntimeError).  The new error class allows callers to detect and handle if there are no backends, without having to catch the ambiguous RuntimeError.

### DIFF
--- a/keyring/backends/chainer.py
+++ b/keyring/backends/chainer.py
@@ -21,9 +21,11 @@ class ChainerBackend(backend.KeyringBackend):
     @classmethod
     def priority(cls):
         """
-        High-priority if there are backends to chain, otherwise 0.
+        If there are backends to chain, high priority
+        Otherwise very low priority since our operation when empty
+        is the same as null.
         """
-        return 10 * (len(cls.backends) > 1)
+        return 10 if len(cls.backends) > 1 else -10
 
     @properties.ClassProperty
     @classmethod
--- a/keyring/backends/fail.py
+++ b/keyring/backends/fail.py
@@ -1,4 +1,5 @@
 from ..backend import KeyringBackend
+from ..errors import NoKeyringError
 
 
 class Keyring(KeyringBackend):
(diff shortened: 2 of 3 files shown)


mitya57 (approved):

ddstreet:
@jaraco does this look ok to you?  I'd like to include pull this into Ubuntu but want to wait for review/merge here first.  @mitya57 are you able to merge?

keyring/backends/chainer.py
```diff
@@ -21,9 +21,11 @@ class ChainerBackend(backend.KeyringBackend):
     @classmethod
     def priority(cls):
         """
-        High-priority if there are backends to chain, otherwise 0.
+        If there are backends to chain, high priority
+        Otherwise very low priority since our operation when empty
+        is the same as null.
         """
-        return 10 * (len(cls.backends) > 1)
```
jaraco, line 26: Apologies for the delay in review, and thanks for the contrib.

I'm not sure -10 is meaningful here. Probably -1 should be used. The reason 10 is used on the positive side is because 1 means minimally recommended, 5 is used for more recommended ones, so 10 means preferably to those.

I may adjust this, but it looks good at first blush. Probably there should be a test to protect against the issue.

jaraco:
Relevant bug is #372 .

keyring/backends/chainer.py
mitya57: I was also in doubt about −10, but then realized that the idea was to have lower priority than `null` backend, which has −1.

Now I see that you resolved this by using `fail.Keyring.priority - 1`, which is indeed better. Thanks!

PR #425, opened 2020-02-29
Use debug logging when loading plugins

### LINKED ISSUE #405: keyring logging at info level as side effect of being imported
This line:

https://github.com/jaraco/keyring/blob/master/keyring/backend.py#L187

generates INFO logs on import. I believe this type of information, given that it's generated as a side effect of import, should be DEBUG to avoid generating output on something as innocuous as import.

If the information must be logged as INFO, I believe this behavior should be deferred until needed, rather.

Imports are hard to order and in any complex environment as effectively non-deterministic. This makes the INFO log output nondeterministic, which makes testing very difficult. Given the core role of keyring in the Python ecosystem, logged output at the INFO level can pop up in surprising ways.

loganrosen:
Closes #405.

### DIFF
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -199,7 +199,7 @@ def _load_plugins():
     entry_points = metadata.entry_points()['keyring.backends']
     for ep in entry_points:
         try:
-            log.info('Loading %s', ep.name)
+            log.debug('Loading %s', ep.name)
             init_func = ep.load()
             if callable(init_func):
                 init_func()


mitya57 (approved):

loganrosen:
@mitya57 Thanks for the approval! Is this something that we can merge?

mitya57:
I was waiting for @jaraco but this is a very straightforward change, so I went ahead and merged it myself.

PR #440, opened 2020-05-25
Use correct XDG default for XDG_CONFIG_HOME

jnsnow:
The default for when `XDG_CONFIG_HOME` is not defined should be
`~/.config` and not `~/.local/share`. This creates another backwards
compatibility break, but one that _check_old_config_root already
handles.


jaraco:
Can you provide a reference as to why `.config` is preferable?

Issue #445, opened 2020-07-07
Kwallet backend fails in environments where sys.argv is not present


AndrewGrossman:
Kwallet backend fails in environments where _sys_ does not have _argv_ available.

In my particular case, I was trying to utilize the Snowflake connector for SQLAlchemy through PostgreSQL Multicorn FDW. I encountered the following stacktrace in my log:

	  File "/usr/lib/python3/dist-packages/multicorn/sqlalchemyfdw.py", line 282, in __init__
	    self.engine = create_engine(url)

	  File "/usr/local/lib/python3.6/dist-packages/sqlalchemy/engine/__init__.py", line 500, in create_engine
	    return strategy.create(*args, **kwargs)

	  File "/usr/local/lib/python3.6/dist-packages/sqlalchemy/engine/strategies.py", line 61, in create
	    entrypoint = u._get_entrypoint()

	  File "/usr/local/lib/python3.6/dist-packages/sqlalchemy/engine/url.py", line 172, in _get_entrypoint
	    cls = registry.load(name)

	  File "/usr/local/lib/python3.6/dist-packages/sqlalchemy/util/langhelpers.py", line 265, in load
	    return impl.load()

	  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2324, in load
	    return self.resolve()

	  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2330, in resolve
	    module = __import__(self.module_name, fromlist=['__name__'], level=0)

	  File "/usr/local/lib/python3.6/dist-packages/snowflake/sqlalchemy/__init__.py", line 8, in <module>
	    from . import snowdialect

	  File "/usr/local/lib/python3.6/dist-packages/snowflake/sqlalchemy/snowdialect.py", line 27, in <module>
	    from snowflake.connector.constants import UTF8

	  File "/usr/local/lib/python3.6/dist-packages/snowflake/connector/__init__.py", line 17, in <module>
	    from .connection import SnowflakeConnection

	  File "/usr/local/lib/python3.6/dist-packages/snowflake/connector/connection.py", line 18, in <module>
	    from .auth import Auth

	  File "/usr/local/lib/python3.6/dist-packages/snowflake/connector/auth.py", line 51, in <module>
	    import keyring

	  File "/usr/lib/python3/dist-packages/keyring/__init__.py", line 3, in <module>
	    from .core import (set_keyring, get_keyring, set_password, get_password,

[...]

PR #446, opened 2020-07-07
Update kwallet backend to support missing sys.argv #445

AndrewGrossman:
#445


sezna:
Hey, I submitted a PR without seeing yours first. This version of the fix still doesn't work in my use case, where `argv` exists but it has a length of `0`, so there's still an index out of bounds error.

jaraco:
Closing in favor of #447.

PR #447, opened 2020-07-15
Use default appid if sys.argv doesn't exist. Ref #445

sezna:
Thanks for all your work with Keyring, I really enjoy the utility. I encountered #445 when running keyring from an environment that didn't provide `argv` . This adds a check that argv exists and has at least some value before accessing it.


jaraco:
This looks good, but it probably deserves tests to cover the use cases. Would you write some tests that monkey-patch `sys.argv` to capture it not existing or being empty, preferably referencing the environments in which these conditions occur?

jaraco:
Wow. Writing a test for this is not straightforward, because the behavior happens at import time, but also because the test suite is organized around an already-constructed keyring.

jaraco:
In 2167d74b1a4fc4da0a11c34e170099292aec5f5f, I wrote a test that captures the expected failure modes, including another that was missed (`sys.argv = None`).

Issue #431, opened 2020-04-09
Add support for password retrieval without username in kwallet backend


jaraco:

muggenhor:
Given that `SecretService` already implements this behavior I think this ticket would be better by referring to KWallet or KDE.

jaraco:
Indeed, that functionality was added in 19.2.0 (Sep 2019). Thanks for clarifying.

PR #459, opened 2020-08-24
feat: support discovering a credential's username on KWallet too

muggenhor:
Implement the get_credential method for the KWallet backend. This now
allows lookup of a credential by 'service' name only, not requiring the
'username'.

Implements #431

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+v21.4.0
+-------
+
+* #431: KWallet backend now supports ``get_credential``.
+
 v21.3.1
 -------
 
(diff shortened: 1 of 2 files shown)


muggenhor:
FYI: the Azure Pipelines' failures look like an intermittent failure (couldn't download Python with nuget). But I cannot find a way to rerun those builds.

jaraco:
I re-ran the tests. It must have been a transient failure, but now the tests are failing because black released a new version, invalidating previously-acceptable code.

jaraco:
Released as v21.4.0.

PR #463, opened 2020-09-01
Fix KWallet.get_credential() regression

MrAnno:
#459 introduced a regression.

Docs:
> get_credential(service, username): Return a credential object stored in
> the active keyring. This object contains at least username and password
> attributes for the specified service, where the returned username may be
> different from the argument.

Prior to this commit, `get_credential()` returned the password as a string, without wrapping the credential into a `SimpleCredential` object.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+v21.4.1
+-------
+
+* #463: Fixed regression in KWallet ``get_credential``
+  where a simple string was returned instead of a
+  SimpleCredential.
+
 v21.4.0
 -------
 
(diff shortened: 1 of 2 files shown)


muggenhor:
@MrAnno thanks for catching my mistake and fixing it :blush:

MrAnno:
@muggenhor My pleasure :) 

IMHO your version of the fix is more straightforward and explicit.

jaraco:
Would be good to get some enhancements to the test suite to capture this guarantee. Just adding some type hints might be sufficient.

jaraco:
As you can see in the above commit, I've added the type hints, but it seems mypy is comfortable with that.

PR #464, opened 2020-09-08
fix: return a SimpleCredential from all return paths of kwallet.get_credential

muggenhor:
Fixes #463


muggenhor:
I failed to notice that #463 was a PR, not just a bug report.

Issue #473, opened 2020-10-28
Importing the keyring module in some Linux configurations can stall for 25 seconds or so


siebenmann:
**Describe the bug**
On some Unix configurations, simply importing the keyring module can cause the Python interpreter for 25 seconds or so. This has been reported on Wayland sessions and in remote logins (eg in pypa/pip#8485). Whether this happens appears to depend on the Linux involved; it appears to occur on Fedora but not necessarily Ubuntu. This may be considered a duplicate of #403, but I think it's a more specific issue.

**To Reproduce**
1. Arrange a Fedora machine with the kf5-kwallet package installed.
2. Remotely log in to the machine when there is no console login session.
3. Run ``time python3 -c "import keyring"`` and observe it takes 25 seconds or so.

**Expected behavior**
This should complete without stalling, especially for 25 seconds or so.

**Environment**

- OS: Fedora 32

```
$ pip list | grep keyring
keyring            21.3.1
$ keyring --list-backends
keyring.backends.fail.Keyring (priority: 0)
keyring.backends.chainer.ChainerBackend (priority: -1)
keyring.backends.SecretService.Keyring (priority: 5)
```
However code inspection suggests that this will happen with the current git version, although I'm not sure.

**Additional context**
The problem happens in backends/kwallet.py, where the ``priority()`` class method attempts to make a DBus connection to the org.kde.kwalletd5 DBus service. On some Linux systems (possibly it requires systemd) with kwalletd installed, this DBus call will trigger an attempt to start kwalletd, which fails but times out very slowly. You can duplicate this with a simple DBus call:
```>>> import dbus
>>> from dbus.mainloop.glib import DBusGMainLoop
>>> bus = dbus.SessionBus(mainloop=DBusGMainLoop())
>>> bus.get_object('org.kde.kwalletd5', '/modules/kwalletd5')
```
The call path of this on import is through core.py's call to ``init_backend()``, which then uses ``backend.by_priority()`` to try to get the priority for every backend, which triggers the call to kwallet.py's ``priority()``.

takluyver:
Perhaps when selecting the backend it would make sense to check if the service is already running (`bus.name_has_owner('org.kde.kwalletd5')`) or is available to be started through D-Bus (`'org.kde.kwalletd5' in bus.list_activatable_names()`). I think those should both be safe to check without the bus trying to launch it.

[...]

Issue #434, opened 2020-04-24
Importing keyring takes two minutes


ep12:
**Describe the bug**
`import keyring` takes approximately 120 seconds, sometimes less, but on average 120 seconds:
![import keyring](https://user-images.githubusercontent.com/26854722/80195873-cde89700-861c-11ea-8c7e-c65e712c1a6e.png)

**To Reproduce**
This is probably not easily reproducable!

**Expected behavior**
Short import time.

**Environment**
- OS: Arch Linux
- WM: [Qtile](https://www.github.com/qtile/qtile)

```
$ pip list | grep keyring
keyring                       21.2.0
$ keyring --list-backends
keyring.backends.chainer.ChainerBackend (priority: -1)
keyring.backends.fail.Keyring (priority: 0)
keyring.backends.kwallet.DBusKeyring (priority: 4.9)
```

**Additional Information**
+ Related: #420 
+ Usually, qtile took less than a second to start. Now it takes approximately two minutes, which is painful considering I can't be productive in those two minutes.
+ I did some research and found out that this insane delay is caused by `import keyring`, because [`keyring` initialises itself](https://github.com/jaraco/keyring/blob/cc30a66c80df9c4c0f2e10e32d3cd6b214e77012/keyring/core.py#L192) when it is imported.
+ A (single!) `dbus` call takes two minutes before it returns (I don't even know if it fails or if it is successfull). But I don't speak _dbus_ so my digging ends here.
+ I `strace`d `qtile`, see [here](https://github.com/qtile/qtile/issues/1657#issuecomment-618004849). A `recvfrom` takes 120 seconds.
+ When using `pip`, the commands I execute do not produce an output for two minutes. That is because `pip` uses `keyring`.
+ Is it really necessary to perform `keyring.core.init_backend` when importing `keyring`?

**Questions**
+ What's wrong with my system?
+ How to fix it?

ep12:
Closed in favor of .

pawamoy:
Must be coming from [kwallet](https://github.com/jaraco/keyring/blob/979f4dd6078b7ee2db3d5ac71dd3e5bfad62331e/keyring/backends/kwallet.py) or [secret storage](https://github.com/jaraco/keyring/blob/6807da5aa404ad36867078a3bf4f3779d0f4135d/keyring/backends/SecretService.py) as they are the only ones using dbus.

ep12:
I updated my system and the problem went the way it came. Silently.

ep12:
[...]

PR #474, opened 2020-11-06
Disable SecretService and KWallet backends if the corresponding names are not available on D-Bus

### LINKED ISSUE #162: Import of keyring takes 25s - or DBusException
```
# time python -m keyring
/usr/bin/python: No module named keyring.__main__; 'keyring' is a package and cannot be directly executed

real    0m25.235s
user    0m0.176s
sys     0m0.033s
```

or 

I sometimes bump into : 

```
  File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 928, in load_modules
    ), fn_, path, desc
  File "/usr/lib/python2.7/dist-packages/salt/cloud/clouds/nova.py", line 100, in <module>
    from salt.utils.openstack import nova
  File "/usr/lib/python2.7/dist-packages/salt/utils/openstack/nova.py", line 11, in <module>
    from novaclient.shell import OpenStackComputeShell
  File "/usr/lib/python2.7/dist-packages/novaclient/shell.py", line 38, in <module>
    import keyring
  File "/usr/lib/python2.7/dist-packages/keyring/__init__.py", line 12, in <module>
    from .core import (set_keyring, get_keyring, set_password, get_password,
  File "/usr/lib/python2.7/dist-packages/keyring/core.py", line 180, in <module>
    init_backend()
  File "/usr/lib/python2.7/dist-packages/keyring/core.py", line 59, in init_backend
    set_keyring(load_config() or _get_best_keyring())
  File "/usr/lib/python2.7/dist-packages/keyring/core.py", line 67, in _get_best_keyring
    keyrings = backend.get_all_keyring()
  File "/usr/lib/python2.7/dist-packages/keyring/util/__init__.py", line 24, in wrapper
    func.always_returns = func(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/keyring/backend.py", line 127, in get_all_keyring
    exceptions=TypeError))
  File "/usr/lib/python2.7/dist-packages/keyring/util/__init__.py", line 35, in suppress_exceptions
    for callable in callables:
  File "/usr/lib/python2.7/dist-packages/keyring/backend.py", line 119, in is_class_viable
    keyring_cls.priority
  File "/usr/lib/python2.7/dist-packages/keyring/util/properties.py", line 22, in __get__
    return self.fget.__get__(None, owner)()
  File "/usr/lib/python2.7/dist-packages/keyring/util/XDG.py", line 18, in wrapper
    return func(*args, **kwargs) * self.multiplier
  File "/usr/lib/python2.7/dist-packages/keyring/backends/SecretService.py", line 32, in priority
    list(secretstorage.get_all_collections(bus))
  File "/usr/lib/python2.7/dist-packages/secretstorage/collection.py", line 158, in get_all_collections
    service_obj = bus_get_object(bus, SECRETS, SS_PATH)
  File "/usr/lib/python2.7/dist-packages/secretstorage/util.py", line 50, in bus_get_object
    return bus.get_object(name, object_path, introspect=False)
  File "/usr/lib/python2.7/dist-packages/dbus/bus.py", line 241, in get_object
    follow_name_owner_changes=follow_name_owner_changes)
  File "/usr/lib/python2.7/dist-packages/dbus/proxies.py", line 248, in __init__
    self._named_service = conn.activate_name_owner(bus_name)
  File "/usr/lib/python2.7/dist-packages/dbus/bus.py", line 180, in activate_name_owner
    self.start_service_by_name(bus_name)
  File "/usr/lib/python2.7/dist-packages/dbus/bus.py", line 278, in start_service_by_name
    'su', (bus_name, flags)))
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 651, in call_blocking
    message, timeout)
DBusException: org.freedesktop.DBus.Error.TimedOut: Activation of org.freedesktop.secrets timed out
```

Are the two problems linked ? Is the 25 seconds linked to a timeout ? Is there a way to disable trying dbus ?

mitya57:
First, check if the name is already taken by some application. If no, check if the service can be activated dynamically.

@takluyver Please review this if you have time. And thanks again for your suggestion!

This fixes #162, #434, #473.

### DIFF
--- a/keyring/backends/SecretService.py
+++ b/keyring/backends/SecretService.py
@@ -35,11 +35,15 @@ class Keyring(KeyringBackend):
             secretstorage.__name__
         if exc:
             raise RuntimeError("SecretStorage required")
-        if secretstorage.__version_tuple__[0] < 3:
-            raise RuntimeError("SecretStorage 3.0 or newer required")
+        if secretstorage.__version_tuple__ < (3, 2):
+            raise RuntimeError("SecretStorage 3.2 or newer required")
         try:
             with closing(secretstorage.dbus_init()) as connection:
-                list(secretstorage.get_all_collections(connection))
+                if not secretstorage.check_service_availability(connection):
+                    raise RuntimeError(
+                        "The Secret Service daemon is neither running nor "
+                        "activatable through D-Bus"
+                    )
         except exceptions.SecretStorageException as e:
             raise RuntimeError("Unable to initialize SecretService: %s" % e)
         return 5
--- a/keyring/backends/kwallet.py
+++ b/keyring/backends/kwallet.py
@@ -46,12 +46,13 @@ class DBusKeyring(KeyringBackend):
             bus = dbus.SessionBus(mainloop=DBusGMainLoop())
         except dbus.DBusException as exc:
             raise RuntimeError(exc.get_dbus_message())
-        try:
-            bus.get_object(cls.bus_name, cls.object_path)
-        except dbus.DBusException:
-            tmpl = 'cannot connect to {bus_name}'
-            msg = tmpl.format(bus_name=cls.bus_name)
-            raise RuntimeError(msg)
+        if not (
+            bus.name_has_owner(cls.bus_name)
+            or cls.bus_name in bus.list_activatable_names()
+        ):
+            raise RuntimeError(
+                "The KWallet daemon is neither running nor activatable through D-Bus"
+            )
         if "KDE" in os.getenv("XDG_CURRENT_DESKTOP", "").split(":"):
             return 5.1
         return 4.9
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,11 @@
+v21.5.0
+-------
+
+* #474: SecretService and KWallet backends are now
+  disabled if the relevant names are not available on
+  D-Bus. Keyring should now be much more responsive
+  in these environments.
+
 v21.4.1
 -------
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,7 @@ include_package_data = true
 python_requires = >=3.6
 install_requires =
 	pywin32-ctypes!=0.1.0,!=0.1.1; sys_platform=="win32"
-	SecretStorage>=3; sys_platform=="linux"
+	SecretStorage>=3.2; sys_platform=="linux"
 	jeepney>=0.4.2; sys_platform=="linux"
 	importlib_metadata >= 1; python_version < "3.8"
 setup_requires = setuptools_scm[toml] >= 3.4.1


keyring/backends/kwallet.py
```diff
@@ -46,6 +46,10 @@ def priority(cls):
             bus = dbus.SessionBus(mainloop=DBusGMainLoop())
         except dbus.DBusException as exc:
             raise RuntimeError(exc.get_dbus_message())
+        if not (bus.name_has_owner(cls.bus_name)
+                or cls.bus_name in bus.list_activatable_names()):
+            raise RuntimeError("The KWallet daemon is neither running nor "
+                               "activatable through D-Bus")
         try:
             bus.get_object(cls.bus_name, cls.object_path)
```
takluyver: I think we need to also remove this bit. As I understand it, the hang occurs when kwallet *is* an activatable name, but attempting to activate it fails. So the check you've added would pass (it would be in `list_activatable_names()`), and then it would try to do this all the same.

To avoid the hang, we have to not attempt to communicate with the service as part of this check. We *could* attempt this only if it's already running, but I think that's overcomplicating it.

keyring/backends/SecretService.py
```diff
@@ -39,6 +44,14 @@ def priority(cls):
...
             with closing(secretstorage.dbus_init()) as connection:
+                proxy = Proxy(message_bus, connection)
+                has_owner, = proxy.NameHasOwner("org.freedesktop.secrets")
+                if not has_owner:
+                    activatable_names, = proxy.ListActivatableNames()
+                    if "org.freedesktop.secrets" not in activatable_names:
+                        raise RuntimeError(
+                            "The Secret Service daemon is neither running nor "
+                            "activatable through D-Bus")
                 list(secretstorage.get_all_collections(connection))
```
takluyver: Ditto this line, for the same reason.

takluyver:
No problem, thank you for doing the work with it. :-)

For some reason I can't see the logs from the Azure pipelines builds that are failing - I don't know if that's a random glitch, or if they're set to private.

keyring/backends/SecretService.py
```diff
@@ -31,6 +31,11 @@ class Keyring(KeyringBackend):
     @properties.ClassProperty
     @classmethod
     def priority(cls):
+        try:
+            from jeepney.bus_messages import message_bus
+            from jeepney.integrate.blocking import Proxy
+        except ImportError:
+            raise RuntimeError("jeepney required")
```
jaraco: Could this functionality be exposed by secretstorage? I don't understand the relationship between secretstorage and jeepney, but keyring seems to have a dependency on both jeepney and secretstorage, even though this change adds the only reference to jeepney.

jaraco:
Right - please disregard the Azure and Appveyor integrations. I'm migrating projects to github actions. I see I haven't done that here yet. I'll do that in master and then a rebase ought to clear things up.

jaraco:
I've rebased to master to fix CI.

mitya57:
Thanks you @takluyver and @jaraco for your comments! I will address them tomorrow.

keyring/backends/SecretService.py
takluyver: > I don't understand the relationship between secretstorage and jeepney

Jeepney handles D-Bus communications, secretstorage uses that to talk to the secrets service. As an analogy, a Github API client could be built on top of requests, which handles HTTP communications.

It would certainly be possible for SecretService to have a `check_available()` function or similar that encapsulated this logic. Up to @mitya57 whether he wants that. :-)

jaraco (approved):
Regardless the solution you come up with, the general approach seems suitable.

keyring/backends/kwallet.py
mitya57: Removed.

keyring/backends/SecretService.py
mitya57: Also removed.

keyring/backends/SecretService.py
mitya57: I added a `check_service_availability()` function to SecretStorage now. I bumped the dependency to 3.2, but the new version is not released yet (I will do it after review).

A question to @takluyver: Do you know if this code can raise any exceptions, provided that the connection was established successfully? Currently keyring only captures `SecretStorageException`, but in that new function I do not convert any exceptions from Jeepney to `SecretStorageException`. I can use my wrapper class `DBusAddressWrapper` to do that, though.

https://github.com/mitya57/secretstorage/blob/master/secretstorage/__init__.py#L84-L90

takluyver: > Do you know if this code can raise any exceptions, provided that the connection was established successfully? 

In theory, it could: the message bus could respond with an error, which Jeepney would turn into a `DBusErrorResponse` exception. Or if it sends back data that's not a valid D-Bus message, or there's a bug in Jeepney's parsing code, there could be all sorts of exceptions. In practice, I think it's unlikely - the message bus is generally reliable (it has to be), and we're not trying to do anything complex with it.

takluyver:
The code changes LGTM now, though I've not tried them locally.

Just to reiterate what I said on #473: this won't altogether eliminate this kind of hang, but I think that delays when calling keyring functions are more palatable than delays on import. I'm working on changes to Jeepney which will include the ability to set a timeout on blocking method calls, which we could use to limit how long it waits.

keyring/backends/SecretService.py
mitya57: Ok, I left the code as is then. If we get any bugs about such exceptions, I will catch them and make that function return False in those cases.

mitya57:
> this won't altogether eliminate this kind of hang, but I think that delays when calling keyring functions are more palatable than delays on import.

I think there are many cases where keyring is imported but isn't going to be used. So for such users the problem will be solved.

> I'm working on changes to Jeepney which will include the ability to set a timeout on blocking method calls, which we could use to limit how long it waits.

I saw that. Thanks!

I have made SecretStorage 3.2.0 release now, so this is ready for merge.

jaraco:
Expect [this build](https://github.com/jaraco/keyring/actions/runs/351355848) to release v21.5.0.

Issue #438, opened 2020-05-07
UTF-8 password decoding should be supported on Windows


ftruzzi:
Hi, and thank you for your work!

Currently, `get_password` and `get_credential` on Windows assume passwords to be encoded in UTF-16. However, I have encountered more than one case where the bytes object was encoded in UTF-8 (possibly by another library) and keyring would crash with a `UnicodeDecodeError`.

I think there should be a way to let the user specify a different encoding, return the bytes object directly or just fall back to UTF-8 before throwing an error. I can work on a PR if this would be a welcome change.

Thanks!

jaraco:
Thanks for reporting the issue.

That's really unfortunate. It does appear as if [the docs are non-specific](https://docs.microsoft.com/en-us/windows/win32/api/wincred/ns-wincred-credentialw) about the interpretation of the CredentialBlob field, that it's "defined by the application." Gah.

I've previously used this library for many years without any portability concerns, but it seems the unconstrained pointer to bytes has invited new consumers of the API to flout the convention and use a different encoding.

I do recall testing this API and using the Microsoft UI to create credentials and those were compatible with keyring, so it's probably some other application that's generating these unconventional forms. It probably doesn't help that MS standardized on UTF-16 early before the rest of industry standardized on UTF-8.

Out of curiosity, if you open one of these credentials in the MS UI, does it support them?

It would also be worthwhile to identify the source of these credentials and at least assess their motivation for choosing a non-conforming encoding. Can you identify the tool that was creating the credentials in UTF-8 encoding?

It may also be worthwhile to find a MS resource to comment as well, but I'd like to start by first refining our understanding of the source of the issue.

ftruzzi:
> Out of curiosity, if you open one of these credentials in the MS UI, does it support them?

I can edit the credential, but I haven't been able to view any password in plain text on my system, the "view" button just doesn't show up. Is this still possible?

> It would also be worthwhile to identify the source of these credentials and at least assess their motivation for choosing a non-conforming encoding. Can you identify the tool that was creating the credentials in UTF-8 encoding?

I saw this happen with two Electron apps: one is Slack, the other is an app which uses https://github.com/atom/node-keytar. I'm not too familiar with this stuff, but if you take a look [here](https://github.com/atom/node-keytar/blob/master/src/keytar_win.cc) (lines 130-136) it looks like they're indeed treating password data as `std::string` instead of wide chars as they do with credentials names.

[...]

PR #482, opened 2020-12-23
Add support for UTF-8 decoding

ftruzzi:
This PR adds support for UTF-8 password decoding as discussed in https://github.com/jaraco/keyring/issues/438

Wondering if we should add a warning when UTF-16 decoding fails? Something like "you can read this password but won't be able to write it back in the same format using this library at this time"?

Thanks!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,11 @@
+v21.3.0
+-------
+
+* #438: For better interoperability with other
+  applications, ``Windows`` backend now attempts to
+  decode passwords using UTF-8 if UTF-16 decoding fails.
+  Passwords are still stored as UTF-16.
+
 v21.2.1
 -------
 
--- a/keyring/backends/Windows.py
+++ b/keyring/backends/Windows.py
@@ -158,9 +158,13 @@ class WinVaultKeyring(KeyringBackend):
             if not res:
                 return None
         try:
-            return SimpleCredential(res['UserName'], res['CredentialBlob'].decode('utf-16'))
+            return SimpleCredential(
+                res['UserName'], res['CredentialBlob'].decode('utf-16')
+            )
         except UnicodeDecodeError:
-            return SimpleCredential(res['UserName'], res['CredentialBlob'].decode('utf-8'))
+            return SimpleCredential(
+                res['UserName'], res['CredentialBlob'].decode('utf-8')
+            )
 
 
 class OldPywinError:
--- a/keyring/backends/Windows.py
+++ b/keyring/backends/Windows.py
@@ -86,7 +86,10 @@ class WinVaultKeyring(KeyringBackend):
         if not res:
             return None
         blob = res['CredentialBlob']
-        return blob.decode('utf-16')
+        try:
+            return blob.decode("utf-16")
+        except UnicodeDecodeError:
+            return blob.decode("utf-8")
 
     def _get_password(self, target):
         try:
@@ -154,7 +157,10 @@ class WinVaultKeyring(KeyringBackend):
             res = self._get_password(service)
             if not res:
                 return None
-        return SimpleCredential(res['UserName'], res['CredentialBlob'].decode('utf-16'))
+        try:
+            return SimpleCredential(res['UserName'], res['CredentialBlob'].decode('utf-16'))
+        except UnicodeDecodeError:
+            return SimpleCredential(res['UserName'], res['CredentialBlob'].decode('utf-8'))
 
 
 class OldPywinError:
--- a/keyring/backends/Windows.py
+++ b/keyring/backends/Windows.py
@@ -1,4 +1,5 @@
 import functools
+import logging
 
 from ..util import properties
 from ..backend import KeyringBackend
@@ -22,6 +23,7 @@ with ExceptionRaisedContext() as missing_deps:
         # force demand import to raise ImportError
         win32cred.__name__
 
+log = logging.getLogger(__name__)
 
 __metaclass__ = type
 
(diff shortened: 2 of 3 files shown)


jaraco:
I've taken the liberty to refactor the behavior into a `DecodingCredential` class, consolidating the behavior.

> Wondering if we should add a warning when UTF-16 decoding fails?

That seems reasonable, especially because a warning could be suppressed by an environment variable.

One additional thing, could you:

- [ ] Create a test to capture the new expectation? The test will probably need to inject a UTF-8 password into the credential store (and make sure it's registered as created in order to be deleted).

ftruzzi:
Thanks! I have added a warning. I wanted to add a test from the beginning, but how can we write a test UTF-8 credential to the credential store if `pywin32-ctypes` doesn't support it?

jaraco:
> Thanks! I have added a warning. I wanted to add a test from the beginning, but how can we write a test UTF-8 credential to the credential store if `pywin32-ctypes` doesn't support it?

Good question. I guess it's not easy. You'd need to find a back door to add the credential, either another executable or perhaps invoking the underlying API directly (through ctypes).

Now that I re-read through the issue, I see we agreed to support reading these before writing them, so perhaps having a test would require at least nominal support for writing them. And clearly, using `Slack` or another electron app is probably not a very portable solution.

So how about let's for now just write a stub for a test with a skip indicating that the functionality isn't available? Something like:

```
@pytest.mark.skip(reason="writing utf-8 isn't yet supported #438")
def test_utf_8_credential(self):
    # todo: save a non-ascii password in UTF-8
    pw = self.keyring.get_password(system, username)
    assert isinstance(pw, str)
```

ftruzzi:
> You'd need to find a back door to add the credential, either another executable or perhaps invoking the underlying API directly (through ctypes).

Thanks! It looks like manually editing the ctypes credential worked :) It can also be a first step towards supporting it officially if you like the hacky solution. Can you take a look at the new test?

Sorry, realizing this just now:
> perhaps having a test would require at least nominal support for writing them.

Up to you then, we can use a stub test for this and consider better support later, as you prefer

jaraco:
> > You'd need to find a back door to add the credential, either another executable or perhaps invoking the underlying API directly (through ctypes).
> 
> Thanks! It looks like manually editing the ctypes credential worked :) It can also be a first step towards supporting it officially if you like the hacky solution. Can you take a look at the new test?

Great!

> > perhaps having a test would require at least nominal support for writing them.
> 
> Up to you then, we can use a stub test for this and consider better support later, as you prefer

No problem. Looks like were close here, and having a working test is best.

I tried applying this patch to ensure that unicode characters get represented correctly:

```diff
diff --git a/tests/backends/test_Windows.py b/tests/backends/test_Windows.py
index 0c63595..645ed89 100644
--- a/tests/backends/test_Windows.py
+++ b/tests/backends/test_Windows.py
@@ -3,7 +3,7 @@ import sys
 import pytest
 
 import keyring.backends.Windows
-from keyring.testing.backend import BackendBasicTests
+from keyring.testing.backend import BackendBasicTests, UNICODE_CHARS
 
 
 @pytest.mark.skipif(
@@ -56,7 +56,7 @@ class TestWinVaultKeyring(BackendBasicTests):
         """
         service = "keyring-utf8-test"
         username = "keyring"
-        password = "utf8-test"
+        password = UNICODE_CHARS
 
         self.set_utf8_password(service, username, password)
         assert self.keyring.get_password(service, username) == password
```

But when I did, the test failed in the final assertion:

```
_____________________________________________ TestWinVaultKeyring.test_read_utf8_password ______________________________________________ 

self = <test_Windows.TestWinVaultKeyring object at 0x000001D80CB63910>

    def test_read_utf8_password(self):
        """
        Write a UTF-8 encoded credential and make sure it can be read back correctly.
        """
        service = "keyring-utf8-test"
        username = "keyring"
        password = UNICODE_CHARS
     
        self.set_utf8_password(service, username, password)
>       assert self.keyring.get_password(service, username) == password
E       AssertionError: assert '雗铗鯗駗ꏗꇗ\uaad7...胑諑럐뷐냐뫐냐苑뻐돐賑뻐뷐' == 'זהכיףסתםלשמו...ръзнакатогьон'
E         - זהכיףסתםלשמועאיךתנצחקרפדעץטובבגןξεσκεπάζωτηνψυχοφθόραβδελυγμίαСъешьжеещёэтихмягкихфранцузскихбулокдавыпейчаюЖълтатадюлябешещастливачепухъткойтоцъфназамръзнакатогьон
E         + 雗铗鯗駗ꏗꇗ꫗鷗鳗꧗黗闗ꋗ郗駗髗꫗ꃗꛗ韗ꟗ꣗ꓗ鏗ꋗꗗ飗闗釗釗鋗鿗뻎뗎菏뫎뗎胏곎뛎觏蓏럎뷎裏藏蟏뿎蛏룎賏臏뇎닎듎뗎믎藏돎볎꿎뇎ꇐ諑뗐裑賑뛐뗐뗐觑金跑苑룐藑볐近돐뫐룐藑蓑胑냐뷐蛑菑럐臑뫐룐藑뇐菑믐뻐뫐듐냐닐译뿐뗐말蟑냐軑雐諑믐苑냐苑냐듐軑믐近뇐뗐裑뗐觑냐臑苑믐룐닐냐蟑뗐뿐菑藑諑苑뫐뻐말苑뻐蛑諑蓑뷐냐럐냐볐胑諑럐뷐냐뫐냐苑뻐돐賑뻐뷐

tests\backends\test_Windows.py:62: AssertionError
```

It appears as if some UTF-8 can be decoded as UTF-16.

jaraco:
Adding just some ascii characters to the password and the test passes, even with the rest of those unicode characters present, so I think that confirms my suspicion that UNICODE_CHARS.encode('utf-8') is also valid UTF-16.

Still, I guess that's an acceptable limitation.

ftruzzi:
Good! I also came to the same conclusion.

Looks like there are no UTF-16 byte order marks in the credentials saved with ctypes so I don't think there's an easy way out of this. Maybe detecting UTF-8 with `chardet` and falling back to UTF-16 would work but that would be an extra dependency and I don't know enough about encodings to know how frequent these collisions are, let alone in passwords!

PR #493, opened 2021-02-25
Fix windows user expansion

nuno-andre:
Thus retrieves the separator from `os.sep`.

```
>>> os.path.expanduser('~/.config')
c:\Users\username/.config

>>> os.path.join(os.path.expanduser('~'), '.config')
c:\Users\username\.config
```

### DIFF
--- a/keyring/util/platform_.py
+++ b/keyring/util/platform_.py
@@ -56,9 +56,9 @@ def _config_root_Linux():
     location.
     """
     _check_old_config_root()
-    fallback = os.path.join(os.path.expanduser('~'), '.config')
     key = 'XDG_CONFIG_HOME'
-    root = os.environ.get(key, None) or fallback
+    home = os.path.expanduser('~')
+    root = os.environ.get(key, None) or os.path.join(home, '.config')
     return os.path.join(root, 'python_keyring')
 
 
--- a/keyring/util/platform_.py
+++ b/keyring/util/platform_.py
@@ -56,7 +56,7 @@ def _config_root_Linux():
     location.
     """
     _check_old_config_root()
-    fallback = os.path.expanduser('~/.config')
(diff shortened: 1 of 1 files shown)


jaraco:
I appreciate that the output looks a little awkward, but it is adequate and functional, and the current code has an elegant simplicity to it that's lost with the proposed change. What's your motivation for making this change?

nuno-andre:
Fear, mainly :sweat_smile:

Not of Python, its path management works like a charm, but of the shell interaction. As an example, this works xplat:

```
home = os.path.expanduser('~')
path = os.path.join(home, '.config')
os.system(f'mkdir {path}')
```
But the current implementation breaks on Windows because it interprets `/.config` like a switch.

(I'm currently working on a network mgmt application that uses [Mitogen](https://mitogen.networkgenomics.com/), which makes heavy use of `os.system`, and I want to forward credentials from several sources through keyring.)

Anyway, I totally agree with the inelegance of the change. I have just fix it, in case you think the change is justified.

mitya57:
It's also possible to use pathlib which should make the code both elegant and correct:

```python
from pathlib import Path
Path.home() / '.config'
```

nuno-andre:
Of course, but that implies a broader change at the library level. This is only a little fix for a very specific use case.

jaraco:
So I was going to implement Dmitry's idea, but when I went to the code, I found this line is in `config_root_Linux`, so this code wouldn't even be reached on Windows. Since Nuno acknowledges there's no actual failure, but fear of it, I'm going to recommend, "don't be afraid," and thanks for sharing.

nuno-andre:
Well, that was a silly joke, the failure would exist if we weren't sanitizing those strings. Anyway, going to `Path` is the best approach, indeed.

PR #507, opened 2021-04-13
Only use external importlib_metadata package when python < 3.8

PureTryOut:
importlib_metadata has been integrated into the Python standard library
in 3.8 and is thus unnecessary for any newer versions.
Let's only use the external importlib_metadata package when actually
running on a Python version that doesn't include it already

This is mostly useful for distributions where there is no need to
package importlib_metadata anymore, and it's future proofing for when
the external package stops being developed anymore

### DIFF
--- a/hook-keyring.backend.py
+++ b/hook-keyring.backend.py
@@ -1,6 +1,11 @@
 # Used by pyinstaller to expose hidden imports
 
-import importlib_metadata as metadata
+import sys
+
+if sys.version_info < (3, 8):
+    import importlib_metadata as metadata
+else:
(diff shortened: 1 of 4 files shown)


PureTryOut:
I... Don't understand why tests are failing. Locally they succeed fine, and I compared the output of the failing `matches` variable on both Python 3.7.10 and on 3.9.4, both were the same. I'm not sure why Python behaves differently on CI.

PureTryOut:
Never mind, closing this after reading https://github.com/pypa/twine/pull/732#issuecomment-809755307

Too bad

Issue #503, opened 2021-03-14
Please revert importlib_metadata improvements.


jaraco:
Based on the failing CI for #502, I think we're just asking for 473465ea386f2b32424f11295e472d8b9ffa90e1 to be reverted... it's a nicer interface for the implementation of `keyring`, but as a consumer of the code I'd prefer to avoid the extra dependency 😕

_Originally posted by @Zac-HD in https://github.com/jaraco/keyring/issues/499#issuecomment-798814962_

jaraco:
That extra dependency will become optional again in Python 3.10 alpha 7.

Keyring is attempting to use the new, cleaner interface for entry points afforded by [importlib_metadata 3.6](https://importlib-metadata.readthedocs.io/en/latest/history.html#v3-6-0) and to avoid the deprecation of the prior interface.

See python/importlib_metadata#284 for motivation for this change.

I'd like for hypothesis to make a similar change. In a future version, I plan to [deprecate the dict-like behavior](https://github.com/python/importlib_metadata/pull/289) and eventually remove it (probably in importlib_metadata 4.x and Python 3.11).

I'd like to recommend that Hypothesis model after what keyring has done here. I think you'll find it's a nicer interface for Hypothesis as well. I'd be happy to help draft the change to enable that behavior, but I'd not presume to do if bumping a dependency (and expanding the number of Python versions on which the backport is necessary) is something your project would wish to avoid. Can you explain why having an added dependency is a problem?

Zac-HD:
Ah, nice!  That's going to be awesome for everyone who uses it 😁 

> I'd like to recommend that Hypothesis model after what keyring has done here. I think you'll find it's a nicer interface for Hypothesis as well. I'd be happy to help draft the change to enable that behavior, but I'd not presume to do if bumping a dependency (and expanding the number of Python versions on which the backport is necessary) is something your project would wish to avoid. Can you explain why having an added dependency is a problem?

Adding a *dev-time* dependency via `keyring` is no problem at all; initially I thought that it was just missing some conditional logic but obviously that's not the case.

[...]

PR #508, opened 2021-04-20
Prefer stdlib importlib.metdata

jonringer:
Importlib_metadata is available as part of the standard library in python >= "3.8".

Noticed it was failing in https://github.com/NixOS/nixpkgs/pull/119819#discussion_r616810205

### DIFF
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -9,7 +9,13 @@ import operator
 
 from typing import Optional
 
-import importlib_metadata as metadata
+try:
+    # python 3.8 and later
+    import importlib.metadata as metadata
+except ImportError:
+    # python 3.7 and earlier
+    import importlib_metadata as metadata
+
 
 from . import credentials, errors, util
 from .util import properties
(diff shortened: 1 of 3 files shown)


mitya57:
Please see https://github.com/jaraco/keyring/issues/503#issuecomment-798973205.

jonringer:
hmm, interesting.

jonringer:
... What's the point of having a backport package if it exports a different API?

Anyway, I'll just a make a note in our packaging about why `importlib-metadata` is still needed for all stable cpython versions.

mitya57:
> ... What's the point of having a backport package if it exports a different API?

It is backport of the API that will be present in Python 3.10.

jonringer:
Can we update https://pypi.org/project/importlib-metadata/ to also mention that the package also backports changes relevant to `python_version < "3.10"`

mitya57:
@jaraco :point_up: can you do it?

jaraco:
Agreed. The readme is unclear. I filed https://github.com/python/importlib_metadata/issues/306 to track.

PR #517, opened 2021-06-21
Make MacOS Error class inherit from KeyringError

Ghostkeeper:
This way it can be caught from outside without using MacOS-specific imports.

The `KeyringError` class specifies that it is the base class of keyring's errors. Well, it turned out it was not the base class of all of them. Keyring was still raising errors that were not subclasses of that `KeyringError` class. This change makes those a subclass of `KeyringError` too.

There is something to be said for that `KeyringDenied` would also have to be a subclass of `KeyringLocked`, and so on. However the descriptions of those errors don't match completely, and this would prevent them from being a subclass of the Error class (unless you want to get into the mess of multiple inheritance). So I think this would be a more safe approach.

With this change, if the user of the library wants to catch all errors related to the keyring, they just have to catch `KeyringError`. Without this change, they would have to make an import on the MacOS API details (which is giving errors for us as well when this is done on Windows, so OS-conditional imports need to be made). This would be difficult to discover for most developers as there is nothing about it in the documentation.

### DIFF
--- a/keyring/backends/macOS/api.py
+++ b/keyring/backends/macOS/api.py
@@ -3,6 +3,7 @@ import ctypes
 import struct
 from ctypes import c_void_p, c_uint16, c_uint32, c_int32, c_char_p, POINTER
 
+from ...errors import KeyringError
 
 sec_keychain_ref = sec_keychain_item_ref = c_void_p
 OS_status = c_int32
@@ -30,7 +31,7 @@ SecKeychainCopyDefault.argtypes = (POINTER(sec_keychain_ref),)
 SecKeychainCopyDefault.restype = OS_status
 
 
-class Error(Exception):
+class Error(KeyringError):
     @classmethod
     def raise_for_status(cls, status):
         if status == 0:


jaraco:
I intended for the API module to be independent of keyring, a layer between the macOS Security API and any other consumer generally. I'd like to consider instead having the keyring backend actually replace these exceptions, rather than make the API module dependent on keyring interfaces. Did you consider that approach? I do worry that such a translation might prove messy, so I'd consider a decorator approach to reduce the amount of repetition and indentation.

Ghostkeeper:
If you would have the MacOS API independent of the rest of Keyring, then the user would need to adjust its code based on the platform to use the different API. Call different functions depending on the OS, catch different exceptions. In order to catch the MacOS exception you need to import the MacOS API module. You can't import the MacOS API module on Linux or Windows (gives an `OSError` that the DLLs don't exist), so you'd need to import the MacOS API module conditionally, and then also conditionally catch the exceptions. The user can't just add `except KeychainDenied` in their code, because on Windows and Linux that class doesn't exist since that module was conditionally imported. So they'd also have to conditionally catch exceptions based on the OS, or replace the exception with a dummy class on the other platforms.

We used the dummy class approach in our application here: https://github.com/Ultimaker/Cura/pull/10010/files
It's not a nice solution.

If the API for MacOS is completely separate, I'd say that you at least need to make it possible to import the MacOS API on other platforms, so that the symbols you need on every platform are still available without adding conditionals in every place the library is used. The user can then just add `except KeychainDenied` and that exception class would be imported on every platform.

But then indeed this PR would be counterproductive. Feel free to close this if you've made that decision :)

jaraco:
I think maybe you misunderstood. I'm not proposing that all clients update to catch a whole array of exceptions. Instead, I'm suggesting that the `macOS.__init__` module take responsibility for providing the consistent Exception interface (based on `KeyringError`), so translating any exceptions from `macOS.api` to keyring exceptions.

Ghostkeeper:
Okay, yeah that would indeed be a friendlier approach towards the user of the library!

Thanks for the discourse :)

A different approach is then desirable to this change, so I'll retract my PR.
We're not working on it in our team any more, now that our workaround is in place, so I won't really get the time to contribute further any more. Good luck and thanks for your efforts to maintain this library!

PR #521, opened 2021-08-07
Add libsecret backend

a-andre:
With the libsecret backend python-keyring can be used in flatpak without opening a hole for org.freedesktop.secrets which allows cross-application password access.

The backend works also outside of flatpak.

([Secret-portal support isn't in the scope of SecretStorage](https://github.com/mitya57/secretstorage/issues/30))


keyring/backends/libsecret.py
```diff
@@ -0,0 +1,103 @@
+from contextlib import closing
```
mitya57: Unused import.

keyring/backends/libsecret.py
```diff
@@ -0,0 +1,103 @@
...
+import logging
+
+from ..util import properties
+from ..backend import KeyringBackend
+from ..credentials import SimpleCredential
+from ..errors import (
+    PasswordDeleteError,
+    ExceptionRaisedContext,
+    KeyringLocked,
```
mitya57, line 10: Unused import.

keyring/backends/libsecret.py
```diff
@@ -0,0 +1,103 @@
...
+)
+
+available = False
+try:
+    import gi
+    gi.require_version('Secret', '1')
+    from gi.repository import Secret
+    available = True
+except ImportError:
```
mitya57: I think the `require_version` call can also raise ValueError.

```suggestion
except (ImportError, ValueError):
```

keyring/backends/libsecret.py
```diff
@@ -0,0 +1,103 @@
...
+        """Get password of the username for the service"""
+        attributes = {
+            "application": self.appid,
+            "service": service,
+            "username": username,
+        }
+        items = Secret.password_search_sync(self.schema, attributes,
+                                            Secret.SearchFlags.UNLOCK, None)
+        for item in items:
+            return item.retrieve_secret_sync().get_text()
```
mitya57: If the collection was locked, and the user refused to unlock it, this will raise

    gi.repository.GLib.Error: secret-error: Cannot get secret of a locked object (2)

keyring/backends/libsecret.py
```diff
@@ -0,0 +1,103 @@
...
+        """Set password for the username of the service"""
+        collection = Secret.COLLECTION_DEFAULT
+        attributes = {
+            "application": self.appid,
+            "service": service,
+            "username": username,
+        }
+        label = "Password for '{}' on '{}'".format(username, service)
+        Secret.password_store_sync(self.schema, attributes, collection,
+                                   label, password, None)
```
mitya57: Similarly, this can raise

    gi.repository.GLib.Error: secret-error: Cannot create an item in a locked collection (2)

This method can also return False, I think this should be handled somehow (e.g. raise PasswordSetError).

keyring/backends/libsecret.py
```diff
@@ -0,0 +1,103 @@
...
+        attributes = {
+            "application": self.appid,
+            "service": service,
+            "username": username,
+        }
+        items = Secret.password_search_sync(self.schema, attributes,
+                                            Secret.SearchFlags.UNLOCK, None)
+        for item in items:
+            removed = Secret.password_clear_sync(self.schema,
+                                                 item.get_attributes(), None)
```
mitya57: Why not use `item.delete_sync()`?

setup.cfg
```diff
@@ -63,6 +63,7 @@ devpi_client =
 keyring.backends =
 	Windows = keyring.backends.Windows
 	macOS = keyring.backends.macOS
+	LibSecret = keyring.backends.libsecret
```
mitya57: I don’t like the capitalization. The official name of libsecret is lowercase: https://wiki.gnome.org/Projects/Libsecret.

```suggestion
	libsecret = keyring.backends.libsecret
```

mitya57 (changes requested):

keyring/backends/libsecret.py
a-andre: Inside flatpak `item` is of type SecretFileItem and this class doesn't implement `delete_sync()`

a-andre:
Thanks for the feedback.

keyring/backends/libsecret.py
```diff
@@ -0,0 +1,139 @@
...
+        """Get password of the username for the service"""
+        attributes = {
+            "application": self.appid,
+            "service": service,
+            "username": username,
+        }
+        try:
+            items = Secret.password_search_sync(self.schema, attributes,
+                                                Secret.SearchFlags.UNLOCK, None)
+        except GLib.Error as error:
```
mitya57, line 64: In my case this method does not raise the unlock error. It returns a list of results anyway, but then trying to call `retrieve_secret_sync()` raises it.

I tested with GNOME Keyring and used Seahorse to lock/unlock the collection. I don’t know how libsecret behaves when it uses its own file storage. Can it raise the error here?

keyring/backends/libsecret.py
```diff
@@ -0,0 +1,139 @@
...
+            quark = GLib.quark_try_string('g-io-error-quark')
+            if error.matches(quark, Gio.IOErrorEnum.FAILED):
+                raise KeyringLocked('Failed to unlock the item!')
+            raise error
```
mitya57: Here and below: it’s better to use just `raise` (without `error`), to preserve the original traceback. See https://stackoverflow.com/a/45986945.

keyring/backends/libsecret.py
```diff
@@ -0,0 +1,139 @@
...
+            "service": service,
+            "username": username,
+        }
+        try:
+            items = Secret.password_search_sync(self.schema, attributes,
+                                                Secret.SearchFlags.UNLOCK, None)
+        except GLib.Error as error:
+            quark = GLib.quark_try_string('g-io-error-quark')
+            if error.matches(quark, Gio.IOErrorEnum.FAILED):
+                raise KeyringLocked('Failed to unlock the item!')
```
mitya57: Here and below: maybe add `from error`?
```suggestion
                raise KeyringLocked('Failed to unlock the item!') from error
```

keyring/backends/libsecret.py
```diff
@@ -0,0 +1,139 @@
...
+        try:
+            stored = Secret.password_store_sync(self.schema, attributes, collection,
+                                                label, password, None)
+        except GLib.Error as error:
+            quark = GLib.quark_try_string('secret-error')
+            if error.matches(quark, Secret.Error.IS_LOCKED):
+                raise KeyringLocked("Failed to unlock the collection!")
+            quark = GLib.quark_try_string('g-io-error-quark')
+            if error.matches(quark, Gio.IOErrorEnum.FAILED):
+                raise KeyringLocked("Failed to unlock the collection!")
```
mitya57: Does `Gio.IOErrorEnum.FAILED` really mean unlock error? How does it differ from `Secret.Error.IS_LOCKED`?

keyring/backends/libsecret.py
```diff
@@ -0,0 +1,139 @@
...
+        Otherwise, it will return the first username and password combo that it finds.
+        """
+        query = {"service": service}
+        if username:
+            query["username"] = username
+        items = Secret.password_search_sync(self.schema, query,
+                                            Secret.SearchFlags.UNLOCK, None)
+        for item in items:
+            username = item.get_attributes().get("username")
+            return SimpleCredential(username, item.retrieve_secret_sync().get_text())
```
mitya57: In this method you did not add error handling.

mitya57 (changes requested):
Better now, but I still have some comments.

Also please fix the style errors from Black which make the CI fail: https://github.com/jaraco/keyring/pull/521/checks?check_run_id=3280102581.

keyring/backends/libsecret.py
a-andre: Yes, `Gio.IOErrorEnum.FAILED` errors are raised when libsecret uses its own file storage (Flatpak scenario) and the collection is locked.

keyring/backends/libsecret.py
a-andre: With `error`, the information is still available using Python 3, but I removed it.

keyring/backends/libsecret.py
a-andre: The IO error is raised when a user cancels flatpak's keyring-unlock request.

keyring/backends/libsecret.py
```diff
@@ -0,0 +1,159 @@
...
+            )
+        except GLib.Error as error:
+            quark = GLib.quark_try_string('g-io-error-quark')
+            if error.matches(quark, Gio.IOErrorEnum.FAILED):
+                raise KeyringLocked('Failed to unlock the item!') from error
+            raise
+        for item in items:
+            username = item.get_attributes().get("username")
+            try:
+                return SimpleCredential(username, item.retrieve_secret_sync().get_text())
```
mitya57: Please fix the remaining black/flake8 error about line too long, and then the CI should pass.

mitya57 (approved):

jaraco (commented):
This approach looks sound.

All that's needed is some background on what this backend is meant to accomplish. Normally, the built-in backends are meant to be backends that are generally useful, and keyring recommends for less common backends to be implemented as an optionally-installed plugin.

What is the role of libsecret? What is the relationship of flatpack to libsecret? How does libsecret relate to SecretService and kwallet backends? What users will benefit from this change? What users will be possibly negatively affected by this change?

I see the priority is set to 5, meaning it's exactly as preferable as SecretService. Do you expect both of these to be viable in the same environment?

mitya57:
@jaraco libsecret is a library from GNOME project whose initial purpose is wrapping the Secret Service D-Bus API, so in this aspect it is very similar to SecretStorage module in Python.

However, starting with version 0.20.0, libsecret also provides a second backend, that is using a local file instead of Secret Service. It is used mainly for flatpak applications, where for security reasons access to Secret Service API is not granted. Flatpak is a technology for app isolation. See https://gitlab.gnome.org/GNOME/libsecret/-/merge_requests/6 for some details.

So having a libsecret backend in Keyring will help flatpak applications to store their passwords.

This new backend is compatible with the existing SecretStorage-based backend when flatpak is not used:
```python
>>> k1 = keyring.backends.libsecret.Keyring()
>>> k2 = keyring.backends.SecretService.Keyring()
>>> k1.set_password('foo', 'bar', 'baz')
>>> k2.get_password('foo', 'bar')
'baz'
```

It may sound like we don't need the SecretStorage backend anymore. But it's not true — SecretStorage and jeepney are pure Python libraries and they can be easily installed from PyPI. It is not true for libsecret: libsecret requires PyGObject which needs to be installed from system packages or built from source.

Regarding the priorities, I don't have a strong opinion but maybe I would lower the priority of new backend from 5 to something like 4.8, just to make sure the behavior is deterministic and not random (4.9 is taken by KWallet).

jaraco:
I notice the tests don't get run. I thought I remembered that at least one Linux backend had its tests run in CI when running under Linux. Perhaps that functionality was lost when migrating from Travis to Github. Is it possible for the tests for libsecret to be run?

a-andre:
With `pygobject; sys_platform=="linux"` added to tox testenv dependencies, the tests are run. But they fail with
`gi.repository.GLib.GError: g-io-error-quark: Cannot autolaunch D-Bus without X11 $DISPLAY (0)`.

mitya57:
You should use `dbus-run-session` to run tests in a headless environment. Look what I do in SecretStorage project:

https://github.com/mitya57/secretstorage/blob/master/.github/workflows/main.yml#L19-L25

I use a mock server there, but having a real gnome-keyring installed should also work.

jaraco (approved):

takluyver:
> It may sound like we don't need the SecretStorage backend anymore. But it's not true — SecretStorage and jeepney are pure Python libraries and they can be easily installed from PyPI. It is not true for libsecret: libsecret requires PyGObject which needs to be installed from system packages or built from source.

Just to note, if people wanted, it should be possible (and maybe an interesting exercise :-) to build a Python library which wraps the [portals secret API](https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.Secret) and stores/retrieves passwords in an encrypted file - either copying whatever libsecret does, or based on one of the existing encrypted file backends ([keyrings.cryptfile](https://pypi.org/project/keyrings.cryptfile/) or [EncryptedKeyring in keyrings.alt](https://github.com/jaraco/keyrings.alt/blob/1b575a2402eb62f9513700931ef3647bcd047d43/keyrings/alt/file.py#L67)). [This example script](https://gitlab.com/takluyver/jeepney/-/blob/master/examples/blocking_choose_file.py) shows how to call a portal API (albeit FileChooser rather than Secret).

It might be trickier to make this easily installable with only pure Python packages, because I think most of the crypto packages need extension modules for speed. But there might be contexts where being independent of the GObject universe is an advantage.

If you do this for real world use, obviously getting the encrypted file right is important - it would probably be easy to do something that works but isn't very secure.

Issue #519, opened 2021-07-30
MacOS implementation uses legacy password storage


GrosQuildu:
Keyring uses methods like `SecKeychainAddGenericPassword` and `SecKeychainFindGenericPassword` in the [MacOS implementation](https://github.com/jaraco/keyring/blob/d5bc1599d13495e23c3f7151765c29478d8d065d/keyring/backends/macOS/api.py#L234). These are part of the "Legacy Password Storage" according to [the Apple documentation](https://developer.apple.com/documentation/security/keychain_services/keychain_items). A newer API would use `SecItemAdd`, `SecItemCopyMatching` etc.

One problem I have with the Legacy API is that I can't use TouchID for authentication and must type a password every time I call `keyring.get_password`. Moving to the newer API should solve that problem.

I am not sure if the newer methods can be used via a `CDLL`.

Would it make sense to migrate to newer methods, or using the Legacy Storage is a hard requirement for the keyring?

jaraco:
I'd be delighted to move to the newer, preferred APIs. There are several other [macOS issues](https://github.com/jaraco/keyring/issues?q=is%3Aissue+is%3Aopen+label%3AmacOS) that would be helped by having access to this more sophisticated API.

I too am unsure if or how one could bind to that API. I would prefer to have a pure Python implementation if possible (through ctypes or cffi), but I wouldn't exclude the possibility of using something like [PyObjC](https://pyobjc.readthedocs.io/en/latest/).

I'd be delighted to help someone think through the challenge and brainstorm on ways to upgrade this interface, but I'll be unlikely to take it on myself anytime soon.

YKdvd:
This got updated with 23.1.0 - #519 was mentioned in the change log, but it was #522 that did it.  This can be closed?

GrosQuildu:
Yes, thanks! Closing.

PR #522, opened 2021-08-12
Update MacOS keychain backend to new APIs

smichaels-tob:
This commit updates the MacOS keychain backend to use the new `SecItemAdd`, `SecItemCopyMatching`, and `SecItemDelete` APIs, using a pure Python implementation with ctypes. This should address issue #519, as well as opening the door for further MacOS fixes — although it still does not appear to be possible to use TouchID to access the keychain, but this seems to be a limitation in Keychain itself.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+v23.1.0
+-------
+
+* #519: macOS backend APIs updated to newer, non-legacy
+  APIs.
+
+
 v23.0.1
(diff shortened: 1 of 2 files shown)


jaraco:
Wow. This is exciting. Thanks for working on it. At first blush, it looks like the tests are only failing due to linter nitpicks. I'll fix those up and see if we can get this merged.

PR #530, opened 2021-09-11
Use Session collection for testing libsecret backend

mitya57:
This is something I needed to apply to make the tests pass for Debian package.

The Session collection is not documented in the specification and is specific to GNOME Keyring, but we do the same for testing SecretStorage backend:

https://github.com/jaraco/keyring/blob/fb37652fabb6fafd52603bf970eeacf54c9e515e/tests/backends/test_SecretService.py#L21

The login collection is not available in a clean environment, which results in this error:

    No such interface “org.freedesktop.Secret.Collection” on object at path /org/freedesktop/secrets/collection/login

### DIFF
--- a/keyring/backends/libsecret.py
+++ b/keyring/backends/libsecret.py
@@ -40,6 +40,7 @@ class Keyring(KeyringBackend):
                 "username": Secret.SchemaAttributeType.STRING,
             },
         )
+        collection = Secret.COLLECTION_DEFAULT
 
     @properties.ClassProperty
     @classmethod
@@ -77,7 +78,6 @@ class Keyring(KeyringBackend):
 
     def set_password(self, service, username, password):
         """Set password for the username of the service"""
-        collection = getattr(self, 'preferred_collection', Secret.COLLECTION_DEFAULT)
         attributes = {
             "application": self.appid,
             "service": service,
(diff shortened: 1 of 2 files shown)


jaraco:
I can't tell - are these `preferred_collection` attributes the same? Why do the tests set them to different values (`/org/freedesktop/secrets/collection/session` and `session`)?

mitya57:
@jaraco libsecret accepts only the alias, while SecretStorage accepts the full path.

Quoting [libsecret documentation](https://people.gnome.org/~stefw/libsecret-docs/libsecret-Password-storage.html#secret-password-store-sync):

> If collection is NULL, then the default collection will be used. Use SECRET_COLLECTION_SESSION to store the password in the session collection, which doesn't get stored across login sessions.

Where [SECRET_COLLECTION_SESSION](https://people.gnome.org/~stefw/libsecret-docs/libsecret-Password-storage.html#SECRET-COLLECTION-SESSION:CAPS) is defined as `"session"`.

Maybe it would be better to use `Secret.COLLECTION_SESSION`, but I wanted to avoid an extra import in tests.

mitya57:
How it works can be seen here:
https://salsa.debian.org/python-team/packages/python-keyring/-/commits/debian/master/

In particular, a job without this patch [failed](https://salsa.debian.org/python-team/packages/python-keyring/-/jobs/1937010) and a job with this patch [passed](https://salsa.debian.org/python-team/packages/python-keyring/-/jobs/1937076).

Issue #504, opened 2021-03-17
Unable to set keyring to OS_X.Keyring


felixgao:
**Describe the bug**
The keyring must be a subclass of KeyringBackend

**To Reproduce**
1. poetry install keyring
2. add the following code
```
import keyring
from keyring.backends import OS_X
keyring.set_keyring(OS_X.Keyring)
```
3. run the script

**Expected behavior**
the keyring is set to OSX's keyring

**Environment**

- OS: OS X 10.15.7

```
$ pip list | grep keyring
keyring                   23.0.0
...
$ keyring --list-backends
keyring.backends.macOS.Keyring (priority: 5)
keyring.backends.fail.Keyring (priority: 0)
keyring.backends.chainer.ChainerBackend (priority: -1)
...
```

**Additional context**
installed using poetry with .venv  created via pyenvs 
python version is 3.6.12

jaraco:
The error message is incorrect. It should say "The keyring must be an instance of KeyringBackend". You need to pass an instance of the backend, i.e. `keyring.set_keyring(OS_X.Keyring())`. Also, be aware that since keyring 22 and later, you should prefer `keyring.backends.macOS` to `keyring.backends.OS_X`.

jaraco:
By the way, if you're on macOS, you don't need to set the keyring. It will be discovered automatically (based on the priorities).

PR #534, opened 2021-10-12
add version to __init__.py to know when to use specific backends.

achapkowski:
Added the product version number in order to know when to use what backends.  This will help addresses the issues like this when backends are deprecated: https://github.com/jaraco/keyring/issues/504 .

### DIFF
--- a/keyring/__init__.py
+++ b/keyring/__init__.py
@@ -7,6 +7,8 @@ from .core import (
     get_credential,
 )
 
+__version__ = "23.2.1"
+
 __all__ = (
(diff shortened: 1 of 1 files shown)


jaraco:
I can do you one better. Use `importlib.metadata.version('keyring')` (or `importlib_metadata.version()` on Python 3.7 and earlier).

Exposing the version in the package is problemmatic, because it requires it to be updated manually, keep in sync with the releases, and can get out of sync. Better is just to rely on package metadata.

PR #542, opened 2021-11-10
Change get_credential to return generic Credential

wwuck:
A keyring backend may want to return SimpleCredential or EnvironCredential


jaraco (approved):
This looks good. Could you also add a new entry to the changelog with an 0.0.1 bump from the last version?

wwuck:
Done

Issue #540, opened 2021-11-09
Support longer passwords in Windows


jaraco:
@jaraco I know this issue is closed, but the Windows limit is getting in our way, too, and I have a possible idea for how to fix it.

Here's the version of Windows I'm on

```
Edition	Windows 10 Pro
Version	21H1
Installed on	‎5/‎1/‎2021
OS build	19043.1320
```

The limit documented by @cvubrugier is the same limit I am experiencing: 1280 max for the password.

Yet some packages use `keyring` to store tokens (JWT) in `keyring`, and these tokens can easily exceed 1280 bytes.

So my idea is to edit `https://github.com/jaraco/keyring/blob/main/keyring/backends/Windows.py`, in particular `_set_password` and `_get_password`, notice the "too big" password, and split it up into as many pieces as necessary, giving each one a unique (and numbered) `TargetName`.  

Since `_set_password` and `_get_password` are the callers of `win32.CredWrite` and `win32.CredRead`, I can transparently split the password up in `_set_password` and then transparently re-join it up in `_get_password` and no other code will be the wiser for it.

I've prototyped this and it works.

Is this a PR you would potentially welcome (or, minimally, indulge?)

_Originally posted by @jrobbins-LiveData in https://github.com/jaraco/keyring/issues/355#issuecomment-963661996_

jaraco:
Yes, I'd welcome such a change. In addition to devising a protocol for splitting and restoring the password, you'll also need a scheme for naming such passwords... and you'll notice that the Windows Keyring already has a hack to support multiple passwords for the same system with different users, so you'll want to be careful these two naming schemes don't interact badly. You'll also want to consider compatibility - what happens when a user stores a long password and then downgrades keyring... or stores a short one and then upgrades keyring? The upgrade path is more important to consider than the downgrade path.

jrobbins-LiveData:
I hope to have time this weekend to work on this.  I will pay attention to the multiple passwords / different users issue.  

Also, I realized I need to pay attention to `_delete_password` and any other interactions that might need to know about the "sharded" entries.

For the upgrade keyring case, I was thinking that the first "shard" would maintain the `TargetName` provided by the user.  If that is the only "shard", then everything should work the same, and a subsequent downgrade of keyring won't be impacted.   

[...]

PR #544, opened 2021-11-16
new approach to larger Windows passwords

jrobbins-LiveData:
@jaraco as discussed in issue #540, I rethought the Windows sharding approach, based on using the actual byte encoding of the password.  In so doing, I realized that by using UTF-16 encoding, we are typically using 2560 bytes for 1280 characters of password text.  

I studied PR #482, which adds a fallback to UTF-8 decoding, but only if the attempt to decode using UTF-16 raises `UnicodeDecodeError`.  I wasn't 100% sure about that, as I think there might be some UTF-8 encoded bytes that decode into UTF-16 erroneously, without raising that exception.

It occurred to me that if we could tag the `keyring` password entry with its actual encoding, we could know for sure, rather than through a heuristic technique.  The [Attributes](https://docs.microsoft.com/en-us/windows/win32/api/wincred/ns-wincred-credential_attributew) member of the [Credential](https://docs.microsoft.com/en-us/windows/win32/api/wincred/ns-wincred-credentialw) structure looked good to me for this tagging.  I added a single value (JSON-encoded for extensibility) with the `Keyword` `keyring_metadata`.   While each entry has up to 64 attributes, and so we could conceivably spread out into several attributes, I figured a single JSON key/value attribute would be simplest to test and maintain. 

In addition to the encoding, I currently have a couple of other key/value pairs, to track the total number of shards along with each shard's index.  This tracking info lets the code know if there are more shards than the initial entry, saving the runtime cost of doing a `get` for the next shard that might not be there (under my previous attempt.)

```python
        dict(
            max_shards=credential["_max_shards"],
            shard_num=credential["_shard_num"],
            encoding="utf-8"
        )
```

I added a test and changed the sizing of one of the tests.  I may have changed the behavior of `test_long_password_nice_error`, and need some guidance on what its correct behavior is meant to be.  The max password size is currently configured to be 2 ** 20 bytes.  The code allows for easy changing of that max size.

I also needed a different interface to the Credentials API, since I wanted to control the encoding  when calling `CredWrite` and also add the JSON attribute.  So I added a `cffi`-based interface directly to `Windows.py`.  This is ugly, but I'm not adept at setuptools and wasn't sure how to properly add a new file to the system.  I also don't know if I updated `setup.cfg` correcly to reflect the new dependency on `cffi`.  I think it is more maintainable to have the code under `keyring`'s roof, rather than relying on the Enthought package.

I think this implementation is better than the last one I tried.  It falls back to the previous decoding approach for older passwords.  It fixes the problem I encountered and documented in the issue #540 [thread](https://github.com/jaraco/keyring/issues/540#issuecomment-968331613).   

I'm eager to get feedback, as I really like `keyring` and have some Windows-based projects that could really benefit from the ability to handle larger passwords.


jrobbins-LiveData:
@jaraco I didn't realize that all the backends get loaded on all the platforms, which caused my code to break things.  Sorry about that!  I made a change and tested on Windows and WSL2 Ubuntu, hoping I've got it right.

jrobbins-LiveData:
@jaraco have you had a chance to look at my Windows backend work?

keyring/backends/Windows.py
```diff
@@ -49,14 +257,19 @@ def value(self):
...
+
+        cred = self["CredentialBlob"]
+
+        # NOTE if we have our metadata, we know how to decode
+        if ATTRIBUTE_KEYWORD in self:
+            return cred.decode(self[ATTRIBUTE_KEYWORD]["encoding"])
+
         try:
-            return cred.decode('utf-16')
+            return cred.decode("utf-16")
```
jaraco: It would be helpful not to make capricious syntactical changes like this. The black config on this project allows for single-quote strings, so it would be easier to read the diff if it didn't have changes unrelated to the goal.

keyring/backends/Windows.py
```diff
@@ -71,7 +284,7 @@ class WinVaultKeyring(KeyringBackend):
     This backend does some gymnastics to simulate multi-user support,
     which WinVault doesn't support natively. See
     https://github.com/jaraco/keyring/issues/47#issuecomment-75763152
-    for details on the implementation, but here's the gist:
+    for details on the implementation, but here"s the gist:
```
jaraco: Yuck. Looks like an over-zealous find/replace.

keyring/backends/Windows.py
```diff
@@ -88,79 +301,152 @@ def priority(cls):
...
+        if not creds or creds["UserName"] != username:
             # It wasn't found so attempt to get it with the compound name
-            res = self._get_password(self._compound_name(username, service))
-        if not res:
+            creds = self._get_password(self._compound_name(username, service))
+        if not creds:
             return None
-        return res.value
+        password = creds.value
+        return password
```
jaraco: Once again, this change looks like an unnecessary stylistic change. It might be a useful change, but it distracts from the issue at hand.

keyring/backends/Windows.py
```diff
@@ -88,79 +301,152 @@ def priority(cls):
...
 
-    def _get_password(self, target):
+    def _get_password_inner(self, target):
         try:
-            res = win32cred.CredRead(
-                Type=win32cred.CRED_TYPE_GENERIC, TargetName=target
-            )
-        except pywintypes.error as e:
-            e = OldPywinError.wrap(e)
```
jaraco: Glad to see this legacy compatibility go away. Probably this change could be made independently and rolled out in advance to ensure that it doesn't break any use-cases.

keyring/backends/Windows.py
```diff
@@ -88,79 +301,152 @@ def priority(cls):
...
-            Persist=self.persist,
-        )
-        win32cred.CredWrite(credential, 0)
+        pwd_utf8_bytes = password.encode("utf-8")
+        pwd_len = len(pwd_utf8_bytes)
+
+        if pwd_len > MAX_PASSWORD_BYTES:
+            raise ValueError(MAX_PASSWORD_BYTES, "_set_password: {pwd_len=} exceeds {MAX_PASSWORD_BYTES=}")
+
+        n = MAX_PASSWORD_BYTES if self._is_init else self._max_password_bytes
```
jaraco: This logic doesn't smell right to me. Under what conditions is `bool(self._is_init) is False`?

keyring/backends/Windows.py
```diff
@@ -88,79 +301,152 @@ def priority(cls):
...
-        credential = dict(
-            Type=win32cred.CRED_TYPE_GENERIC,
-            TargetName=target,
-            UserName=username,
-            CredentialBlob=password,
-            Comment="Stored using python-keyring",
-            Persist=self.persist,
-        )
-        win32cred.CredWrite(credential, 0)
+        pwd_utf8_bytes = password.encode("utf-8")
```
jaraco: Probably the biggest concern is this new default encoding. The reason keyring defaulted to UTF-16 was because that's the encoding that other tools (except Electron) appear to use UTF-16 by default. See #438 for more details. I'd like for the stored passwords to be interoperable where possible (i.e. a password stored by keyring should be readable by other programs). Obviously, they won't be interoperable if they are sharded, but this change alters the default encoding as well. And unfortunately, there's no standard here, just a popular convention. I don't think it will be advisable to change the default encoding to utf-8 until there's a substantial move to support that format in other tools. This change would also be backward incompatible in that it would make new passwords incompatible with older keyrings prior to support for UTF-8 decoding.

.gitignore
```diff
@@ -57,3 +57,6 @@ docs/_build/
 # PyBuilder
 target/
 
+# VSCode
+.vscode
+
```
jaraco: Nitpick: It looks like at some point, someone bullied me into including a whole load of crap in this file, giving rise to the temptation for each person to customize this project to suit their needs. In my opinion, it's an anti-pattern to keep a copy of the aggregate of all development environments in every project. I personally keep common ignores in a global ignores file and reference that from my git system ignores, such that it's unnecessary to supply these for each and every project I maintain.

But since this project has a grab-bag of ignores from #184, I can't fault you for including your preferences. In the future, this will get reverted, so I recommend creating a global .gitignore. Maybe I'll write a blog about it explaining a better way.

keyring/backends/Windows.py
```diff
@@ -1,46 +1,254 @@
...
 
 
+# region standalone replacement for win32cred
 with ExceptionRaisedContext() as missing_deps:
+    import cffi
+
+    if sys.platform != "win32":
+        raise EnvironmentError(f"Windows backend requires sys.platform = 'win32' but {sys.platform=}")
+    else:
+        ffi = cffi.FFI()
```
jaraco: I would suggest moving all of the CFFI stuff into an `api` module, similar to what the `macOS` backend does.

tests/backends/test_Windows.py
```diff
@@ -58,9 +29,25 @@ def test_read_utf8_password(self):
...
+
+    def test_long_password_too_long_nice_error(self):
+        try:
+            self.keyring.set_password('system', 'user', 'x' * (2**20 + 1))
+            self.keyring.delete_password('system', 'user')
+        except ValueError as e:
+            if e.args[0] == 2**20:
+                assert True
+            else:
+                assert False
```
jaraco: Eww. Just use `assert e.args[0] == 2**20`.

jrobbins-LiveData:
@jaraco I didn't realize that Python 3.6.15 support was required -- my bad.  I've fixed my code and tested on Ubuntu with Python 3.6.15.

jaraco (changes requested):
I have a lot of nitpicks and one serious concern (change in default encoding). I'm also still not confident that the new sharding strategy is only invoked when necessary, so that's something I'll want to verify. Have a look at my first runthrough and I can take another look.

keyring/backends/Windows.py
jrobbins-LiveData: Agreed.  I'm quite new at working on open source (although have been programming for nearly 50 years), and don't know a lot.  Is there a document for `keyring` giving guidance for a new developer?  If so happy to read it.  In any case, I'll revert to single-quote strings.

keyring/backends/Windows.py
jrobbins-LiveData: Yep, my bad.

keyring/backends/Windows.py
jrobbins-LiveData: This is actually a meaningful change.  The previous code wan't checking if the the new username was different from the old username.  My change fixes #545.  I just now read your comment on #545, and  I agree that a unit test is in order.   

I really don't know the "convention" in this kind of thing.  My thinking was "well, I have the hood up, might as well fix this while I'm here." Perhaps this fix should be made in a separate PR?

keyring/backends/Windows.py
jrobbins-LiveData: I don't think I understand your comment.  Are you saying that this change should be made in a separate Pull Request?  I'm not sure how to do that.

keyring/backends/Windows.py
jrobbins-LiveData: As commented in `__init__`,

```
        # prevent sharding during self._discover_max_password_bytes
        self._is_init = True
        self._max_password_bytes = self._discover_max_password_bytes()
        self._is_init = False
```

`_discover_max_password` returns the "measured" maximum length that fits in a Credentials Manager password record. `_discover_max_password` uses `_set_password` to "binary search" for the maximum size.  During that search, before we have the measured result, the search sets an arbitrary large `MAX_PASSWORD_BYTES` to bound the seach.  After we've measured the true "max", `self._is_init` is set to `False`, and the code simply uses the measured result (`self._max_password_bytes`).

[EDIT]
I am rewriting `_discover_max_password()` to use its own version of `_set_password()`  to remove the need for the confusing `self._is_init` flag.

keyring/backends/Windows.py
jrobbins-LiveData: I'm fine sticking with the UTF-16 encoding.  I only switched because I noticed that, for token storage, we were wasting half the bytes in the record.  A typical JWT fits in a single record under UTF-8, but spills to two sharded records in UTF-16.  In that case (a typical JWT), either way other programs on Windows aren't going to be able to read it, unless they support the proposed attribute scheme. 

But, yes, a password that would fit under UTF-16 would become non-backwards compatible with my code, so I will switch back to UTF-16.

[EDIT] Upon further reflection, perhaps our new encoding rule could be "encode in UTF-16 unless the password wouldn't fit, in which case use UTF-8".  This seems consistent with the concerns you articulated, as a password that _could_ be interoperable _would_ still be interoperable.  An advantage of this rule is that JWTs might very fit in one record, and thereby not require the added cost of sharding.  Yet any password that fits when UTF-16 would still be compatible with other tools and older keyrings.  What do you think?

.gitignore
jrobbins-LiveData: I know that my ignorance is possibly irritating, but I have no idea what a "global .gitignore" is.  I will look that up and read about it.  I only updated the project's .gitignore to make sure I wasn't messing something up by using VSCode.  I meant no harm, and hope you can chalk it up to my ignorance (which I am trying to correct!!)

[EDIT] ok, Windows global `.gitignore` is now in place, and I'll remove my change to the project's `.gitignore`.

keyring/backends/Windows.py
jrobbins-LiveData: I will study that backend.  I definitely want it in a separate module.  I wasn't sure what the right way was to do that.

tests/backends/test_Windows.py
jrobbins-LiveData: I'm embarrassed.  Of course your way is a lot nicer!

keyring/backends/Windows.py
jrobbins-LiveData: Agreed.

jrobbins-LiveData:
@jaraco I pushed a bunch of changes meant to implement your suggestions.  Aside from any remaining blunders on my part (which I will fix once pointed out!), I think the material topics are:

1. validating my change to [set_password](https://github.com/LiveData-Inc/keyring/blob/917ab760a6c8bf903594486744b17f026ecd32c9/keyring/backends/Windows.py#L172), meant to only do the "collision produces new compound name" logic if the user name is different -- I need help creating a unit test showing the behavior described in issue #545 and then showing that it is fixed.  By observation, I can see that my PR version on my machine no longer has extraneous near-duplicate entries in Credentials Manager.
2. validating that the code only shards if the password is too long when encoded in UTF-16

Happy Thanksgiving!  Looking forward working through these topics.

jrobbins-LiveData:
@jaraco I ran `tox` on Windows, and found myself having to workaround some kind of `pytest-flake8` / `flake8` compatibility problem which obscured the `flake8` findings until I replaced a line of code in  `lake8\formatting\base.py`, changing

```
sys.stdout.buffer.write(output.encode() + self.newline.encode())
```

to 

```
print("output.encode()")
```

so I could see what flake8 had to say about my code.  I then fixed those findings.

keyring/backends/Windows.py
jaraco: Yes. That's what I was suggesting. No problem. I committed the change in 0426944, which I'll merge into this PR.

jaraco:
> I found myself having to workaround some kind of `pytest-flake8` / `flake8` compatibility problem

You've probably found by now, the issue is tracked in https://github.com/tholo/pytest-flake8/issues/81.

keyring/backends/Windows.py
jaraco: I'd planned to merge the change into this PR, but found I don't have permission to push to your fork. It's a good idea when making a pull request to (a) Create a branch (other than main) in your own fork for the change and (b) grant maintainers of the upstream fork access to that branch. You can pull my merge commit using `git pull https://github.com/jaraco/keyring LiveData-Inc/main`.

keyring/backends/Windows.py
jaraco: No problem. Glad for your contribution and happy to help. There's no keyring-specific guidance, but as a general rule, and especially early on when working with a project, try to stick to changes that directly impact the issue at hand. See [this blog](https://blog.jaraco.com/how-to-write-perfect-pull-request/) for some guidance I wrote years ago, which is mostly still relevant.

tests/backends/test_Windows.py
```diff
@@ -58,8 +29,22 @@ def test_read_utf8_password(self):
         username = "keyring"
         password = "utf8-test" + UNICODE_CHARS
 
-        self.set_utf8_password(service, username, password)
```
jaraco, line 61: I'm afraid this change no longer adequately sets a UTF-8 password.

tests/backends/test_Windows.py
```diff
@@ -58,8 +29,22 @@ def test_read_utf8_password(self):
...
+        self.keyring.delete_password('system', 'user')
+
+    def test_long_password_too_long_nice_error(self):
+        try:
+            self.keyring.set_password('system', 'user', 'x' * (2 ** 20 + 1))
+            self.keyring.delete_password('system', 'user')
+        except ValueError as e:
+            assert e.args[0] == 2 ** 20
+        except Exception:
+            assert False
```
jaraco: Use `pytest.raises` to assert a specific exception.

tests/backends/test_Windows.py
```diff
@@ -58,8 +29,22 @@ def test_read_utf8_password(self):
         username = "keyring"
         password = "utf8-test" + UNICODE_CHARS
 
-        self.set_utf8_password(service, username, password)
+        self.keyring.set_password(service, username, password)
         assert self.keyring.get_password(service, username) == password
+        self.keyring.delete_password(service, username)
+
+    def test_long_password_nice_error(self):
```
jaraco, line 36: The naming of this function is misleading. "nice error" doesn't make sense if no error occurs. Oh, I see there was previously a test that had this same behavior. Probably in a past life, credentials were limited to 512 characters (or thereabouts) and the test would assert that the error was a nice one. Looks like that test was added in ddfe3324 as part of #355. It looks like I started writing a test that would assert a nice error message when the password was too long, but then stopped short when I couldn't even replicate the issue with a 512 character password. I probably should have just kept going until I reached the limit, then ensured that the resulting error message was nicer.

keyring/backends/Windows.py
jrobbins-LiveData: Clearly I have a lot to learn.  Sorry for my repeated mistakes.  I wish I had read your [blog post](https://blog.jaraco.com/how-to-write-perfect-pull-request/) before I had started!

I will pull your merge commit, and grant you access.  I'm assuming (out of sheer ignorance) that it is too late for me to create a branch etc?

jaraco (commented):
> I also needed a different interface to the Credentials API, since I wanted to control the encoding when calling `CredWrite` and also add the JSON attribute. So I added a `cffi`-based interface directly to `Windows.py`. This is ugly, but I'm not adept at setuptools and wasn't sure how to properly add a new file to the system. I also don't know if I updated `setup.cfg` correcly to reflect the new dependency on `cffi`. I think it is more maintainable to have the code under `keyring`'s roof, rather than relying on the Enthought package.

I'd like to revisit this consideration. While I do like the idea of not being dependent on another implementation, it means keyring is essentially creating its own re-implementation of the Enthought package and adopting the burden that goes along with that (separate maintenance burden, diverging implementations, loss of re-use).

If we step back, I'd like to capture an issue for why pywin32-ctypes is inadequate (probably upstream in that project) and consider solving that issue first before chosing to fork a new implementation. I do think it's plausible that the upstream implementation isn't flexible enough for keyring's needs due to its legacy association with pywin32 and the interfaces it provided. I would like to shed _that_ legacy.

If we decide to adopt a unique implementation, I'd like to consider whether cffi is the best approach. Adding a dependency on cffi adds a _binary_ dependency for keyring, meaning anyone needing keyring cannot install from source without having a build environment set up. Historically, that was a much bigger problem before the advent of eggs/wheels and their common adoption in PyPI. This issue was recently [reported in twine](https://github.com/pypa/twine/issues/837#issuecomment-979275593) when users of Alpine Linux were required to install keyring and by extension cffi, but could not.

Finally, while I believe the sharding implementation here is sound enough (and provably so with the proposed additional tests), I'd like to consider a refactoring that would separate the concerns (sharded vs. unsharded). In particular, I imagine a `SimpleKeyring` that implements `WinVaultKeyring` as currrently implemented and then a specialized subclass `ShardedKeyring` that implements the sharding behavior. Even if the `SimpleKeyring` is disabled (priority -1), it would provide a means to easily test without the sharding behavior and also make it clearer what functionality is pertinent to that concern. That separation of concerns is apparent in this diff, but it's going to be much harder to reason about going forward.

To summarize, here's how I'd like to move forward:

1. Determine if pywin32-ctypes can support the necessary interfaces herein and at least capture any shortcomings.
2. Port the existing backend to an API that supports the necessary interfaces (perhaps an updated version of pywin32-ctypes or maybe a bespoke implementation). Cut a release to get early feedback on any regressions related to that change.
3. Implement the ShardedKeyring as a specialized subclass or mix-in, encapsulating the sharding logic in one class.

This incremental approach, while it may take longer, should give us better visibility into the effects of different changes.

keyring/backends/Windows.py
jaraco: What you did was pretty good. It's a little awkward that the `macOS` capitalization made it to `windowsOS` (which isn't a thing), but that's easily corrected. The main thing I would do differently is to keep everything for Windows in a single package, like so:

```
$ git mv keyring/backends/Windows.py keyring/backends/Windows/__init__.py
$ git mv keyring/backends/windowsOS/api.py keyring/backends/Windows
$ git rm keyring/backends/windowsOS/__init__.py
```

And then change the import to use `from . import api`.

.gitignore
jaraco: In [this section](https://blog.jaraco.com/skeleton/#ignoring-artifacts), I've documented my strategy for `.gitignore`.

jrobbins-LiveData: Thank you for that specific reference.  I also found this [article](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) which I'm assuming (based on looking at `~/.gitconfig`) uses

```
git config --global core.excludesfile ~/.gitignore_global
```

to edit `.gitconfig` for the user.  Your approach of directly editing `.gitconfig` seems both easier and also helped me learn how `git` is globally configured.

tests/backends/test_Windows.py
jrobbins-LiveData: I believe some other contributor supplied that limit of 512, using, logically enough, the Microsoft documentation.  But that documentation is outdated (and not updated by Microsoft) regarding their new limit of 2560 bytes (1280 16-bit encoded characters.)  Which is why 512 worked.

keyring/backends/Windows.py
jaraco: No worries, and don't be so hard on yourself ;)

> I'm assuming (out of sheer ignorance) that it is too late for me to create a branch etc?

It's never too late. Except maybe for this specific pull request. But since elsewhere, I've recommended to step back and approach the problem differently, it will be a good opportunity to work with branches. Here's what I'd recommend (I'm going to assume you're using a command-line client, but you can find similar operations in a GUI client):

1. First, you want to save the work you've already done in this PR (in your main branch) by giving it a branch name. Make sure you've committed any changes you want to keep or discard any changes you do not. Then, while still on your main (with the latest changes pulled), just run `git branch feature/windows-shard-pr544`. That will create a new branch with its head pointing to the same commits as your current main. That's not normally how one creates a branch, but it's a good way to create a branch without checking out that branch.
2. Since you're still on `main`, reset it to the state of upstream. If you don't already have an `upstream` remote configured, you should configure one with `git remote add upstream https://github.com/jaraco/keyring`. Then make sure you have the commits from upstream with `git fetch upstream`. Then reset main with `git reset --hard upstream/main`.
3. Now you're ready to create a new branch for new development. Use `git checkout -b feature/windows-shard-take-2` (you can call the branch whatever you want; I just suggest names like I might use).
4. Finally, after you've committed some changes to your branch and you want to create a pull request, you first need to push that branch to your fork. Assuming your fork is the "origin", while on the branch, you do that with `git push --set-upstream origin` (I think; you may also need to pass the name of the branch).

I know this branching stuff can get confusing, so I do recommend grabbing an SCM visualization GUI like Sourcetree (free) or Sublime Merge ($) or PyCharm ($). It helps me visualize the changes that are happening to different branches of development and their state across local and remote repos. If you love the command line, you may also be interested in the [gh tool](https://github.com/cli/cli) as it helps facilitate a lot of github operations (such as creating pull requests) from the command-line.

Hope that helps you get started on your next branch!

tests/backends/test_Windows.py
jrobbins-LiveData: I broke that when I removed `pywin32-ctypes`.  Here's the thing though, I think purely by chance, the test used to work due to the choice of an odd (rather than even) length `password`.  If you change

```
password = "utf8-test" + UNICODE_CHARS
```

to

```
password = "utf8-test0" + UNICODE_CHARS
```

at https://github.com/jaraco/keyring/blob/04269447ae7d30d3e31b9122fcd0e0358c87eb51/tests/backends/test_Windows.py#L59, you'll see the default `'utf-16'` decode "work", but not with the desired results.

The only way I could think to make the desired results work was to add an `attribute` to the Credentials Manager entry, explictly stating the encoding uses for the password.  

I should have done a better job documenting my concern, and not leaving this test saying one thing and doing another.

jrobbins-LiveData:
Re your comment above, taken a point at a time

> To summarize, here's how I'd like to move forward:
> 
> 1. Determine if pywin32-ctypes can support the necessary interfaces herein and at least capture any shortcomings.
> 2. Port the existing backend to an API that supports the necessary interfaces (perhaps an updated version of pywin32-ctypes or maybe a bespoke implementation). Cut a release to get early feedback on any regressions related to that change.
> 3. Implement the ShardedKeyring as a specialized subclass or mix-in, encapsulating the sharding logic in one class.

1.  Agreed.   Is that something you would like me to try to do, or will you be doing that?  I'm fine either way.  Ideally, I'd like to coordinate the nature of the request being made of [pywin32-ctypes](https://github.com/enthought/pywin32-ctypes) with you.

2. Understood, and agreed, although I'd like to discuss with you the existing backend's approach to "fallback to utf-8", which, I'm afraid, doesn't accomplish the desired goal.  I realize this is a separate concern, and sorry for lumping it together with other concerns. I think I've been fixated on resolving this (alleged) issue by means of the `attribute` approach shown in PR #544. 

It turns out that the approach of "try utf-16, and if it doesn't fit, encode in utf-8 and mark the record with an attribute stating the encoding" actually fixes the issue that I turned to sharding for in the first place, since doubling the effective character count for a JWT makes my JWT actually fit in one record -- no sharding needed.

3. Given that my immediate need is met by the ability to encode in utf-8 if the utf-16 encoding doesn't fit, and further given that this approach allows one to reliably determine the encoding at decode time (which doesn't help us recognize third-party utf-8, unfortunately, but I fear this is a known, hard problem), perhaps step 3 in your action plan might want to stop here, with the use of an encoding attribute, and forego the sharding altogether.  This may prove short-sighted, as the sharding approach allows for passwords longer than 2560 bytes.  So the sharding approach could be added as a step 4 to your plan.  Or we could stick to a single step 3, working like the PR I wrote.  What do you think?

jaraco:
> 1. Is that something you would like me to try to do, or will you be doing that?

I'd be delighted if you could take this on. I can provide guidance when needed. If it's on me to do it, I'm not sure when it'll get the necessary attention.

> So the sharding approach could be added as a step 4 to your plan. Or we could stick to a single step 3, working like the PR I wrote. What do you think?

Agreed - let's defer sharding for as long as possible. If we can find a safe way to encode to UTF-8 and essentially buy 100% more space, let's do that first.

However, let's be cautious. I'm a little skeptical of the attribute-based approach due to portability/interoperability concerns. Ideally, keyring should be compatible with other consumers and generators of credentials (something that's quite difficult to test).

jrobbins-LiveData:
Regarding

> However, let's be cautious. I'm a little skeptical of the attribute-based approach due to portability/interoperability concerns. Ideally, keyring should be compatible with other consumers and generators of credentials (something that's quite difficult to test).

I'd like to try to understand your concerns about using the credential's attributes.  I agree that testing is difficult, and ends up being only as good as the "uses in the wild" that we know about, and, in my case certainly I know very little about that!

First, re "other consumers" of credentials.  I think the most conservative approach would be to not generate an attribute if the password is encoded in utf-16.  A "keyring classic Windows entry" would continue to have a utf-16 encoded password, and we wouldn't add an attribute.  (NOTE: this is **not** as coded currently in my PR.) This is meant to describe how keyring works now, without my changes, and so, presumably, wouldn't represent any added specific concerns regarding "other consumers" when adding the new utf-8 feature.

Re "other generators" of credentials: assuming an interface to the credentials which lets us see the attribute(s), if any, keyring would identify its special attribute, if present, and, by design, this attribute would only be there is written there by keyring.   (This is currently coded in my PR.)  So keyring would be happy with or without the attribute.  Without the attribute, keyring would continue to assume that the password is utf-16 encoded.  (We still need to address correctly "other generated credentials" encoded in utf-8.)

The only change introduced by the above (more conservative approach) would be to credentials where the password is not currenly storable by keyring.  I don't think there is any new interoperability or compatiblility issue, since we'd only be adding entries that otherwise aren't there to be interoperated with in the first place.  No existing user of keyring on Windows would see any changes to their keyring-generated credentials.  

It is true that were there to be a third party consumer of a keyring-generated credential, that consumer would not recognize the utf-8 encoding of a password that, up to this hypothesized new release, it couldn't consume.  So, conceivably, a third-party app would now see an entry, and be unprepared for the utf-8 encoding therein.  So, technically, this is a new risk.  I'm not sure how to evaluate how likely this case is, or what the repercussions would be for that third party app.  Hopefully, the app would behave no differently from being handed a wrong password, but I don't know.

There's also the risk that an app wouldn't want to see attributes whatsoever, despite being part of the API definition.  I have no way of evaluating that risk, either. 

So, yes, in the case of encoding a password that would otherwise not fit in utf-8, it is conceivable that a third-party app would have issues, either with the utf-8 encoding, or with the presence of an attribute.

If these risks are unacceptable, we could bail altogether on both utf-8 **and** on attributes, and simply shard based on a naming convention.  While I have subjective bias against this (cuz I like the efficiency of utf-8 and think the attribute scheme is tidier than a naming convention),  putting that bias aside,  this could be coded using the existing `pywin32-ctypes` API (that's a big win!) and would maintain the existing utf-16 encoding.   

The risk to a third-party consumer would seemingly be reduced in this case to the risk of picking up the partial first shard's contents as if it were the whole password.  But that's akin to having the wrong password, which can happen for many reasons, so doesn't seem like a big new risk we'd be introducing.

This argues for backing away from utf-8 and from attributes, and thereby not needing a new interface.  Sharding, in this light, seems "simpler" to me.  What do you think?

jaraco:
I regret the delay. Been busy and just now getting a chance to look into this proposal again.

> This argues for backing away from utf-8 and from attributes, and thereby not needing a new interface. Sharding, in this light, seems "simpler" to me. What do you think?

Sharding does seem to be the safest approach, though I also agree less elegant than transitioning to UTF-8 encoding. Such a transition _may_ in fact be inevitable (if other applications start producing/expecting utf-8, irrespective of attributes).

For that reason, my preference would be to migrate to UTF-8. I also agree that a sharding strategy can be handled independently. I'll leave it to you to decide how you'd like to proceed, subject to the advice above (introduce necessary API changes first, separately, from feature changes).

PR #549, opened 2021-11-25
Add equality operator to EnvironCredential

wwuck:
It would be useful for using EnvironCredential in unit tests if it has the equality operator implemented.

### DIFF
--- a/keyring/credentials.py
+++ b/keyring/credentials.py
@@ -39,6 +39,15 @@ class EnvironCredential(Credential):
         self.user_env_var = user_env_var
         self.pwd_env_var = pwd_env_var
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, EnvironCredential):
+            return NotImplemented
+
+        return (
+            self.user_env_var == other.user_env_var
+            and self.pwd_env_var == other.pwd_env_var
+        )
+
     def _get_env(self, env_var):
         """Helper to read an environment variable"""
         value = os.environ.get(env_var)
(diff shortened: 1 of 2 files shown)


jaraco:
> It would be useful for using EnvironCredential in unit tests if it has the equality operator implemented.

Can you say more about where this would be useful? I don't see any unit tests for EnvironCredential.

keyring/credentials.py
```diff
@@ -39,6 +39,15 @@ def __init__(self, user_env_var, pwd_env_var):
...
         self.pwd_env_var = pwd_env_var
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, EnvironCredential):
+            return NotImplemented
+
+        return (
+            self.user_env_var == other.user_env_var
+            and self.pwd_env_var == other.pwd_env_var
+        )
```
jaraco: I think this can be more concisely represented by:

```suggestion
        return vars(self) == vars(other)
```

I don't think it's necessary to return NotImplemented, as Python simply transforms that into False anyway, and it seems a vanishingly small chance that anything other than an EnvironCredential would have the same internal state and end up being compared with an EnvironCredential for equality.

wwuck:
Apologies, I missed the review before you merged this.  I've pushed the initial code up to github at https://github.com/wwuck/keyrings.envvars but I can't push to TestPyPI or PyPI until https://github.com/pypa/warehouse/issues/10030 is fixed.

Issue #554, opened 2021-11-29
Windows backend UTF-8 detection doesn't work


jrobbins-LiveData:
**Describe the bug**
UTF-8 detection in [Windows.py](https://github.com/jaraco/keyring/blob/0f0faf1243c6e53c9258bf47eaf6aa5b55514112/keyring/backends/Windows.py#L45) relies on an initial attempt to decode as UTF-16 raising `UnicodeDecodeError`.   The unit test, by using an odd-character length UTF-8 test string, shows the detection apparently working, but substitution of an even-length test string shows it doesn't work.

[I found this issue while working on improving the Windows backend password maximum length of 1280 characters, and wanted to capture it as a separate issue.]

**To Reproduce**
1. Copy the existing [test_read_utf8_password](https://github.com/jaraco/keyring/blob/0f0faf1243c6e53c9258bf47eaf6aa5b55514112/tests/backends/test_Windows.py#L53) test to a new method, e.g. `test_read_even_utf8_password`

```
    def test_read_even_utf8_password(self):
        """
        Write a UTF-8 encoded credential and make sure it can be read back correctly.
        """
        service = "keyring-utf8-test"
        username = "keyring"
        password = "utf8-test0" + UNICODE_CHARS

        self.set_utf8_password(service, username, password)
        assert self.keyring.get_password(service, username) == password
```

2. Run the unit tests on Windows
3. See error

>         assert self.keyring.get_password(service, username) == password
> E       AssertionError: assert '瑵㡦琭獥ぴ雗铗鯗駗ꏗꇗ\...胑諑럐뷐냐뫐냐苑뻐돐賑뻐뷐' == 'utf8-test0זה...ръзнакатогьон'
> E         - utf8-test0זהכיףסתםלשמועאיךתנצחקרפדעץטובבגןξεσκεπάζωτηνψυχοφθόραβδελυγμίαСъешьжеещёэтихмягкихфранцузскихбулокдавыпейчаюЖълтатадюлябешещастливачепухъткойтоцъфназамръзнакатогьон
[...]

Issue #545, opened 2021-11-18
Password in Windows credentials is not updated on next `set_password` call


wl2776:
**Describe the bug**
New entry is created instead of updating existing password.

**To Reproduce**

1. Open cmd.exe and set password with CLI utility
    ```
    C:\>keyring set service username
    Password for 'username' in 'service': <enter the password>
    ```

2. Open Credential manager and observe a new entry:
    ![image](https://user-images.githubusercontent.com/4774284/142363885-08a7dc78-7e8e-413e-bd1f-20db1173cbc5.png)

3. Update password. Open cmd.exe again and set new password with CLI utility for the same service and user name
    ```
    C:\>keyring set service username
    Password for 'username' in 'service': <enter another password>
    ```

4. Refresh Credential manager's window and see two entries: `service` and `username@service`

    ![image](https://user-images.githubusercontent.com/4774284/142364254-8b5b29a9-a5f9-4fa8-a508-35e8f6d3b44e.png)

**Expected behavior**
Second call for `keyring set service username` should modify existing entry instead of creating new one.

**Environment**

- OS: Windows 10 Pro with all recent updates.

```
C:\> pip list | grep keyring
keyring                    23.2.1

C:\> keyring --list-backends
keyring.backends.Windows.WinVaultKeyring (priority: 5)
keyring.backends.chainer.ChainerBackend (priority: -1)
keyring.backends.fail.Keyring (priority: 0)
```

**Additional context**
The same behavior is observed in call to `keyring.set_password` in python scripts.

jrobbins-LiveData:
I've done some work on the Windows backend  (see pull request #544) that, I believe, fixes this issue.

jaraco:
That's interesting. Probably a bug in how the Windows backend hacks around Windows' limitation to only store one password per site. Glad to hear the pending PR might address that. Would be nice to get a unit test to capture the expectation.

jaraco:
Oh, man. I think it's worse than reported above. Now that I look at the implementation, it appears as if `get_password` [gives preference to the credential with the simple name](https://github.com/jaraco/keyring/blob/04269447ae7d30d3e31b9122fcd0e0358c87eb51/keyring/backends/Windows.py#L99), but `get_credential` gives [preference to the credential with the compound name](https://github.com/jaraco/keyring/blob/04269447ae7d30d3e31b9122fcd0e0358c87eb51/keyring/backends/Windows.py#L165).

[...]

PR #555, opened 2021-12-05
"take-2": new Windows backend, ctypes-based, separate class which shards via Attributes

jrobbins-LiveData:
@jaraco, here's my fresh effort to implement an improved Windows backend, to the plan you outlined:

1. Determine if pywin32-ctypes can support the necessary interfaces herein and at least capture any shortcomings.
2. Port the existing backend to an API that supports the necessary interfaces (perhaps an updated version of pywin32-ctypes or maybe a bespoke implementation). Cut a release to get early feedback on any regressions related to that change.
3. Implement the ShardedKeyring as a specialized subclass or mix-in, encapsulating the sharding logic in one class. 

Re 1, I contacted [pywin32-ctypes](https://github.com/enthought/pywin32-ctypes/issues/105), and heard back that it would take a week or two to hear back from them.  I wrote the code that I would like to have from that project.  It is in `keyring/backends/Windows/api.py`.  I implemented it, this time, using only `ctypes`, and not `cffi`.

Re 2, I moved the existing `backends/Windows.py` to `backends/Windows/__init__.py`, and ported it to use `api.py`.  

Re 3, I implemented `class WinVaultAttributesKeyring(WinVaultKeyring)` to isolate the new sharding logic.

In order to study use-cases of `Credential Manager`, I implemented `CredEnumerate`, which let me more readily dump out all the credentials in my `Credential Manager`.  I learned that, apparently, Microsoft had the same issue we have, in needing more room for JWTs.  

Their sharding solution uses Attributes, rather than creating multiple entries.  I liked that much better, and so the sharding logic in this PR uses Attributes.  I'm including a redacted  version of that dump here, which includes the first few bytes of `CredentialBlob` from each entry.  I sure hope I redacted enough!!!
 
[redacted_credentials2.txt](https://github.com/jaraco/keyring/files/7656549/redacted_credentials2.txt)

While the introduction of the use of Attributes would be new to `keyring`, I restrict their use to the "_password won't fit and we raise  `1783` error_" case.  Passwords that fit are treated the same as in shipping `keyring`, so the risk introduced should be small.

I added a test illustrating the problem raised in https://github.com/jaraco/keyring/issues/554.  `def test_read_even_length_utf8_password(self)`

I also fixed the extra duplication of entries raised in https://github.com/jaraco/keyring/issues/545.  But I didn't know how to fix `get_credentials`, per [the observation you made](https://github.com/jaraco/keyring/issues/545#issuecomment-981121347).  

**What's left?:**
I don't understand the `priority` system 100%, and need some guidance there.  But I did add some tests explicitly targeting the new sharding backend.

I'd like to resolve the `get_credentials` part of the `compound_name` issue.  I don't understand the algorithm well enough to fix it myself.  In general, the idea that I call `set_password` with one `service`, and, later, the system renames it, yet transparently finds it for me, is pretty confusing to me.  I would benefit from more info about what problem is being solved.  Is it that the `keyring` model is that only the `(service, username)` is meant to be unique?  

My credentials file shows that only `keyring` users (and presumably other `pywin32` or `pywin32-ctypes` users) are using `UTF-16`.  Many of the credentials appear to be various 8-bit encodings, and typically do not appear to be Unicode encodings at all.  If my `Credential Manager` data is indicative, I don't think we are helping interoperability by imposing `UTF-16`.  Rather, I think we simply inherited this behavior from `pywin32`.  I'd love to discuss what we can do about this, without disrupting existing users.

### DIFF
--- a/keyring/backends/Windows/__init__.py
+++ b/keyring/backends/Windows/__init__.py
@@ -1,26 +1,13 @@
 import logging
 
-from ..util import properties
-from ..backend import KeyringBackend
-from ..credentials import SimpleCredential
-from ..errors import PasswordDeleteError, ExceptionRaisedContext
+from ...util import properties
+from ...backend import KeyringBackend
+from ...credentials import SimpleCredential
+from ...errors import PasswordDeleteError, ExceptionRaisedContext
 
 
 with ExceptionRaisedContext() as missing_deps:
-    try:
-        # prefer pywin32-ctypes
-        from win32ctypes.pywin32 import pywintypes
-        from win32ctypes.pywin32 import win32cred
-
-        # force demand import to raise ImportError
-        win32cred.__name__
-    except ImportError:
-        # fallback to pywin32
-        import pywintypes
-        import win32cred
-
-        # force demand import to raise ImportError
-        win32cred.__name__
+    from . import api as win32cred
 
 log = logging.getLogger(__name__)
 
@@ -48,7 +35,12 @@ class DecodingCredential(dict):
         """
         Attempt to decode the credential blob as UTF-16 then UTF-8.
         """
+        # If the credential blob was already decoded, e.g.
+        # by CredReadFromAttributes, simply return it.
         cred = self['CredentialBlob']
+        if isinstance(cred, str):
+            return cred
+
         try:
             return cred.decode('utf-16')
         except UnicodeDecodeError:
@@ -65,7 +57,7 @@ class WinVaultKeyring(KeyringBackend):
     WinVaultKeyring stores encrypted passwords using the Windows Credential
     Manager.
 
-    Requires pywin32
+    Requires Windows
 
     This backend does some gymnastics to simulate multi-user support,
     which WinVault doesn't support natively. See
@@ -87,7 +79,7 @@ class WinVaultKeyring(KeyringBackend):
         If available, the preferred backend on Windows.
         """
         if missing_deps:
-            raise RuntimeError("Requires Windows and pywin32")
+            raise RuntimeError("Requires Windows")
         return 5
 
     @staticmethod
@@ -109,26 +101,32 @@ class WinVaultKeyring(KeyringBackend):
             res = win32cred.CredRead(
                 Type=win32cred.CRED_TYPE_GENERIC, TargetName=target
             )
-        except pywintypes.error as e:
-            if e.winerror == 1168 and e.funcname == 'CredRead':  # not found
+        except OSError as e:
+            if e.winerror == 1168:  # not found
                 return None
             raise
+
         return DecodingCredential(res)
 
-    def set_password(self, service, username, password):
+    def set_password(self, service, username, password, encoding='utf-16-le'):
         existing_pw = self._get_password(service)
         if existing_pw:
-            # resave the existing password using a compound target
             existing_username = existing_pw['UserName']
-            target = self._compound_name(existing_username, service)
-            self._set_password(
-                target,
-                existing_username,
-                existing_pw.value,
-            )
-        self._set_password(service, username, str(password))
+            # resave the existing password using a compound target
+            # Fixes part of https://github.com/jaraco/keyring/issues/545,
+            # but get_credentials also needs to be fixed to search in the same
+            # order as get_password.
+            if existing_username != username:
+                target = self._compound_name(existing_username, service)
+                self._set_password(
+                    target,
+                    existing_username,
+                    existing_pw.value,
+                    encoding=encoding,
+                )
+        self._set_password(service, username, str(password), encoding=encoding)
 
-    def _set_password(self, target, username, password):
+    def _set_password(self, target, username, password, encoding):
         credential = dict(
             Type=win32cred.CRED_TYPE_GENERIC,
             TargetName=target,
@@ -137,7 +135,7 @@ class WinVaultKeyring(KeyringBackend):
             Comment="Stored using python-keyring",
             Persist=self.persist,
         )
-        win32cred.CredWrite(credential, 0)
+        win32cred.CredWrite(credential, 0, encoding=encoding)
 
     def delete_password(self, service, username):
         compound = self._compound_name(username, service)
@@ -153,11 +151,13 @@ class WinVaultKeyring(KeyringBackend):
     def _delete_password(self, target):
         try:
             win32cred.CredDelete(Type=win32cred.CRED_TYPE_GENERIC, TargetName=target)
-        except pywintypes.error as e:
-            if e.winerror == 1168 and e.funcname == 'CredDelete':  # not found
+        except OSError as e:
+            if e.winerror == 1168:  # not found
                 return
             raise
 
+    # TODO https://github.com/jaraco/keyring/issues/545
+    # check non-compound_name first?
     def get_credential(self, service, username):
         res = None
         # get the credentials associated with the provided username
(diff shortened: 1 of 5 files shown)


jrobbins-LiveData:
@jaraco any feedback on this rewrite?

keyring/backends/Windows/__init__.py
```diff
@@ -109,26 +101,32 @@ def _get_password(self, target):
...
+            # order as get_password.
+            if existing_username != username:
+                target = self._compound_name(existing_username, service)
+                self._set_password(
+                    target,
+                    existing_username,
+                    existing_pw.value,
+                    encoding=encoding,
+                )
+        self._set_password(service, username, str(password), encoding=encoding)
```
jaraco, line 127: I'm confused by this comment. It's saying that by re-saving the existing password using a compound target, it fixes part of 545, yet re-saving the existing password was already part of the prior change. Moreover, "get_credentials also needs to be fixed" feels like a misplaced TODO. Additionally, the comment doesn't say anything about what the code is meant to do, only has a reference to a reported issue.

By reading the issue, I now think the comment is only about the "if existing_username" check. The problem with putting the comment here is it becomes interleaved with the other comment, making it difficult to tell what comment goes with which parts of code. Since this change adds a whole new branch of code, it may be time to consider a refactor.

At the very least, I'd move the "resave" comment back to where it was (at the beginning of the `if existing_pw` branch and then move the rest of the comment into the `if existing_username` branch and rephrase to indicate its essential purpose, something like, "Only set the compound target when usernames match. Ref #545."

keyring/backends/Windows/api.py
```diff
@@ -0,0 +1,273 @@
...
+
+def CredRead(Type, TargetName):
+    pcred = pointer(CREDENTIALW())
+    _CredReadW(TargetName, Type, 0, byref(pcred))
+    credential = pcred.contents.asdict()
+    _CredFree(pcred)
+    return credential
+
+
+def CredReadFromAttributes(Type, TargetName, Credential=None, encoding='utf-8'):
```
jaraco, line 157: nitpick: I'd probably name this using PEP 8 conventions. The reason for using the capital camel case is to match the names/parameters of the Windows API. In this case, the functionality is not part of the Windows API, but is a helper function, so probably should follow the naming of other helper functions.

keyring/backends/Windows/api.py
```diff
@@ -0,0 +1,273 @@
...
+        accum = b''
+        for i, (key, attr) in enumerate(attrs.items()):
+            expected_key = '{}:{}'.format(ATTRIBUTE_KEYWORD, str(i))
+            if key == expected_key:
+                accum += string_at(attr['Value'], attr['ValueSize'])
+            else:
+                break
+
+        Credential['CredentialBlob'] = accum.decode(encoding)
+        Credential['CredentialBlobSize'] = len(Credential['CredentialBlob'])
```
jaraco, line 173: It seems a little odd to me to be utilizing the API structures to embed the altered values, values that wouldn't be allowed in the structure. I think I'd rather see a different approach where the encoding/decoding strategy sits as a layer between the Windows API and the keyring API, and less intertwined with the Windows API.

I can provide an example later.

tests/backends/test_Windows.py
```diff
@@ -10,56 +8,96 @@
     not keyring.backends.Windows.WinVaultKeyring.viable, reason="Needs Windows"
 )
 class TestWinVaultKeyring(BackendBasicTests):
-    def tearDown(self):
-        # clean up any credentials created
-        for cred in self.credentials_created:
-            try:
-                self.keyring.delete_password(*cred)
-            except Exception as e:
-                print(e, file=sys.stderr)
```
jaraco, line 19: I don't understand what's going on here. This functionality seems to have been removed without justification. This code is there for a reason, preventing the pollution of a user's own keychain when running tests. Why was it removed?

jaraco (changes requested):
Overall, this is looking _really_ good. I quite like the direction it's gone and it's homing in on a sound solution.

I still worry there's just too much going on. This one change affects several different behaviors (introducing new keyring, switch to local ctypes API, changes existing pw checks, introduce encoding, introduces tests for broken encoding, ...). Although I think the final implementation may look very much like what you have here, it's just too much changing for one commit.

I'd much rather see this change as a series of incremental changes, each addressing more selective concerns, either as separate, independent PRs or as separate commits if very closely related.

For example: PR1 could implement simply switching from pywin32ctypes to a native ctypes implementation. That change could be reviewed, accepted, and released so that its effect could be proven and any unexpected regressions could be addressed before accepting more changes that depend on that change. Then PR2 could implement the sharding strategy (based on PR1). PR3 could fix the existing password checks and be independent of PR1 and PR3.

How would you feel about that approach?

keyring/backends/Windows/__init__.py
jrobbins-LiveData: Agreed.

keyring/backends/Windows/api.py
jrobbins-LiveData: When I was writing it, I was trying to imagine the function I wish was already in the API, and wrote the one I wish was there.  But that's all inside my head.  I agree that it should be named as a helper, not as a faux member of the real API.

keyring/backends/Windows/api.py
jrobbins-LiveData: Agreed (see my previous comment for how I got there.)

tests/backends/test_Windows.py
jrobbins-LiveData: When I ran my tests, I noticed that this function wasn't getting called.  Furthermore, the `cleanup` function in the underlying `BackendBasicTests` _was_ getting called (seemingly accomplishing the same goal, although without the `try...except`.

Looking at the `pytest` documentation, I see that there is an optional `teardown` (lowercase `"d"`) method that a test class can provide.  My guess is that this function was intended to be named `teardown` and was mistakenly named `tearDown`, which made it invisible to `pytest`.

Since the base class `BackendBasicTests` `cleanup` seemed to clean up all the unit test password in my run, I don't know that we need this one. 

I should have documented why I yanked it -- sorry about that!

jrobbins-LiveData:
I like the idea of a phased set of pull requests.  The only drawback is that I am inexperienced both with git and with tox etc on Linux.  So I am getting bogged down in the mechanics.   

I could really use some instructions as to how to install on Ubuntu so I can test.

Also, I don't know how to create new pull requests into my account without clobbering the existing cloned repo. 

Other than these mechanical issues, I'd be happy to break it all apart into 3 PRs.  Maybe you or someone else could undertake PR3 (and fix the typo with `tearDown` in the existing unit test for the Windows backend, since both of those issues are, as you say, unrelated to addressing the problem of accepting longer passwords on Windows?

[EDIT]
It appears that adding

```
import sys
if not sys.platform.startswith('win'):
    collect_ignore.append('keyring/backends/Windows/api.py')
```

to `conftest.py` would fix the tox/pytest crash on ubuntu etc.  Is that the correct fix?

jaraco:
> I could really use some instructions as to how to install on Ubuntu so I can test.

For me, the easiest way to test on Ubuntu is to download Docker. I have a "tox-multipy" image I use to test Python applications. From the keyring source folder:

```
docker run -it -v $(pwd):/src jaraco/multipy-tox
# cd src
# $TOX_WORK_DIR='/tmp/tox'
# tox
```

> Also, I don't know how to create new pull requests into my account without clobbering the existing cloned repo.

Oh, that must be really frustrating. I'd recommend to get a nice repo Gui like VSCode or Sourcetree or PyCharm or Sublime Merge (maybe in that order) and use that to manage your repo. It will make it easier to discover operations (like creating a branch and pushing it to your Github repo) and also visualize your changes. I'd also recommend reading through some of the [tutorial resources on using Github](https://docs.github.com/en/get-started/quickstart/hello-world), which talk about creating branches in Github.

For your current code, I'd recommend to create a branch for it.

```
git branch windows-take-2-ctypes
```

Then reset your main to match upstream and reset to that state.

```
git checkout main
git fetch upstream
git reset --hard upstream/main
```

Note, you may or may not have a remote called "upstream". If not, you may need to create one first.

```
git remote add upstream https://github.com/jaraco/keyring
```

I hope that helps. I would recommend spending some time getting familiar with branching and submitting PRs from branches as that will make your life immensely easier when it comes to contributing these changes.

Thanks for your patience and perseverance on this effort!

jaraco:
Closing as stale. Feel free to revive this effort at any time.

Issue #561, opened 2022-02-11
libsecret headless check?


davegaeddert:
**Describe the bug**

I'm trying to debug an issue in another library and it seems to stem from the libsecret backend failing in `get_password`:
https://github.com/jaraco/keyring/blob/eb1a04a42617131a35ed8b6089f8889e6054bcb1/keyring/backends/libsecret.py#L65-L68

The environment is a GitHub Action (ubuntu), and the error is `g-io-error-quark: Cannot autolaunch D-Bus without X11 $DISPLAY (0)
`. Here's a link to the action output: https://github.com/dropseed/workhorse/runs/5160584552?check_suite_focus=true#step:3:263

I found some other discussion about headless, but I guess what I'm wondering is whether there could be some kind of check in libsecret `priority` (or elsewhere) that would prevent it from using that keyring in the first place?
https://github.com/jaraco/keyring/blob/eb1a04a42617131a35ed8b6089f8889e6054bcb1/keyring/backends/libsecret.py#L45-L52

I'm not familiar enough with the details, but I see some stuff that looks a little like that in the kwallet backend:
https://github.com/jaraco/keyring/blob/eb1a04a42617131a35ed8b6089f8889e6054bcb1/keyring/backends/kwallet.py#L40-L58

Or is there some way that it is their responsibility to use `get_keyring` differently?
https://github.com/python-poetry/poetry/blob/7cc684981983963dc202e1a249a4b66667b468bd/src/poetry/utils/password_manager.py#L86-L95

Thanks!

jaraco:
Ideally, the backend should not have a non-zero priority if it is unlikely to work when queried. I'd welcome a fix to make it so.

davegaeddert:
@jaraco do you have any suggestions for a fix? It's not something I know much about. The only thing I could come up with was to actually do a dummy password search during `priority` and catch the specific error... which doesn't seem like a great idea?

takluyver:
Probably the best approach is for the `priority` method to actually try connecting to D-Bus and catch the same error there. This is what the SecretStorage and kwallet backends already do:

https://github.com/jaraco/keyring/blob/d1e37a5efb56ab4b805b721847c87fb94b9ca57b/keyring/backends/SecretService.py#L41-L46

https://github.com/jaraco/keyring/blob/d1e37a5efb56ab4b805b721847c87fb94b9ca57b/keyring/backends/kwallet.py#L45-L52

But I don't know exactly what to call in libsecret to do something equivalent.

[...]

PR #564, opened 2022-02-13
Test libsecret in CI

davegaeddert:
I don't have a fix yet for #561, but this change to the tests does surface it in CI. Previously it was skipped in every matrix because `import gi` usually doesn't work due tox / Python install / virtual environments.

I think the solution is to make `libsecret.Keyring.viable` be false in this kind of environment, but I haven't figured out a good way to do that yet (sounds like there are several other scenarios where libsecret should still work). Just wanted to push the failing test part in case anyone else is thinking about it!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,6 +5,7 @@ on: [push, pull_request]
 jobs:
   test:
     strategy:
+      fail-fast: false
       matrix:
         python:
         - 3.7
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,10 +14,15 @@ jobs:
         - ubuntu-latest
         - macos-latest
         - windows-latest
+        include:
+        - python: "os"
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -16,18 +17,23 @@ jobs:
...
+        if: ${{ matrix.python != 'os' }}
         uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python }}
       - name: Install tox
         run: |
           python -m pip install tox
       - name: Run tests
-        run: tox
+        run: tox ${{ matrix.toxargs }}
```
jaraco, line 36: I'd prefer if some setting in the environment would be sufficient, rather than having to plumb this setting through to the invocation. I've filed https://github.com/tox-dev/tox/issues/2473 to request that feature.

davegaeddert: Sounds good

PR #570, opened 2022-04-12
Fix keyring.util.platform_.config_root on Windows

Darsstar:
Added missing underscore

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+v????
+-------
+
+* The correct config root is now used on Windows.
+
 v23.5.0
 -------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,4 +1,4 @@
-v????
+v23.5.1
 -------
 
 * The correct config root is now used on Windows.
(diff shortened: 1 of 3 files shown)


tests/util/test_platform.py
```diff
@@ -0,0 +1,28 @@
...
+    _config_root_Windows,
+    _data_root_Linux,
+    _data_root_Windows,
+)
+
+
+def test_platform_Linux():
+    # rely on the Github Actions workflow to run this on different platforms
+    if platform.system() != "Linux":
+        return
```
mitya57: I think it's better to skip test, not return (which will make it appear as successful when it has not run). You can use this decorator for example:
https://docs.pytest.org/en/latest/how-to/skipping.html#id1

Darsstar: Done!

mitya57 (approved):

CHANGES.rst
```diff
@@ -1,3 +1,8 @@
+v????
```
jaraco:
```suggestion
v23.5.1
```

Issue #369, opened 2019-01-03
On Mac, is it possible to use any keychain-db other than login/default?


RichardBronosky:
Because of corporate IT, my login keychain is not secure. (It can be accessed by people (corp IT) other than me.) I have move my "service" entry to a keychain file that is exclusive to me, but I can't figure out how to get keyring to use it. If I tell "Keychain Access" to make it the default, it will use it, but, I'd rather not leave it like that.

Please advise.

jaraco:
I don’t believe it’s readily possible today. I believe there’s even an open request to support other common keychains like Internet passwords and not just Generic credentials.

I’d certainly entertain an enhancement to support this need. Perhaps you would be interested in digging deeper into the implementation to devise and demonstrate what changes would be needed to support this feature?

RichardBronosky:
For what it is worth, https://github.com/99designs/aws-vault manages to use a dedicated keychain on macOS rather than the default. I don't know if there is some magic that being a signed binary an advantage to a golang app over a python app.

ev0rtex:
While it's not exposed or abstracted to a usable cross-platform concept you technically _can_ get it to use a different macOS keychain if you use it via the python API (see the [keyring/backends/OS_X.py](https://github.com/jaraco/keyring/blob/7915afdaae9dc664a2ba1ffd91abe64efecc0d46/keyring/backends/OS_X.py#L17) file)...but it's a total hack that involves modifying the class `keychain` value:
```py
import sys
import keyring
from pathlib import Path

if __name__ == '__main__':
    kr = keyring.get_keyring()
    kr.keychain = str(Path.home() / 'Library' / 'Keychains' / (sys.argv[1] + '.keychain-db'))
    print(kr.get_password('service', 'username'))
```

jaraco:
That approach, although hacky, seems suitable for a one-off approach. I'd be open to supporting an environment variable or some other means of soliciting alternate keychains. The real challenge lies in how the interface is common across platforms, so platform-specific inputs aren't available through the interface. Each keyring backend implementation needs to solicit its configuration from the environment. Regardless, I welcome suggestions and contributions.

PR #572, opened 2022-05-02
allow keychain to be specified in macOS

a350h:
so this is my first ever contribution to a public repo and hoping that I didn't miss anything (git is a royal PITA). Very open to suggestions and critique. 

I did add a test for this feature, but it is hard to do from a user agnostic point of view. Because its a mac only feature, I created a new test body just for Mac.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -72,6 +72,14 @@ The basic usage of keyring is pretty simple: just call
     >>> keyring.get_password("system", "username")
     'password'
 
+When using keyring with the macOS Keychain, you can pass the full file path of
+a keychain-db to customize which keychain keyring utilizes. Note: the keychain must
+already exist::
+
+    >>> import keyring
+    >>> path = "/Users/foobar/Library/Keychains/Keys.keychain-db"
+    >>> keyring.set_password("system", "username", "password", path)
+
 Command-line Utility
 --------------------
 
--- a/keyring/backends/macOS/__init__.py
+++ b/keyring/backends/macOS/__init__.py
@@ -1,4 +1,5 @@
 import platform
+import os
 from os.path import exists
 
 from ...backend import KeyringBackend
@@ -18,7 +19,7 @@ class Keyring(KeyringBackend):
     """macOS Keychain"""
 
     # keychain = os.environ.get('KEYCHAIN_PATH')
-    # "Path to keychain file, overriding default"
+    # "Path to keychain file, overriding default. function param "
 
     @properties.ClassProperty
     @classmethod
@@ -39,6 +40,8 @@ class Keyring(KeyringBackend):
         if keychain != "":
             if not exists(keychain):
                 raise KeyringLocked("Can't open specified keychain")
+        elif os.environ.get('KEYCHAIN_PATH') != "":
+            keychain = os.environ.get('KEYCHAIN_PATH')
 
         try:
             api.set_generic_password(keychain, service, username, password)
@@ -54,6 +57,8 @@ class Keyring(KeyringBackend):
         if keychain != "":
             if not exists(keychain):
                 raise KeyringLocked("Can't open specified keychain")
+        elif os.environ.get('KEYCHAIN_PATH') != "":
+            keychain = os.environ.get('KEYCHAIN_PATH')
 
         try:
             return api.find_generic_password(keychain, service, username)
@@ -71,6 +76,8 @@ class Keyring(KeyringBackend):
         if keychain != "":
             if not exists(keychain):
                 raise KeyringLocked("Can't open specified keychain")
+        elif os.environ.get('KEYCHAIN_PATH') != "":
+            keychain = os.environ.get('KEYCHAIN_PATH')
 
         try:
             return api.delete_generic_password(keychain, service, username)
--- a/keyring/backends/macOS/__init__.py
+++ b/keyring/backends/macOS/__init__.py
@@ -1,5 +1,5 @@
 import platform
-import os
+from os.path import exists
 
 from ...backend import KeyringBackend
 from ...errors import PasswordSetError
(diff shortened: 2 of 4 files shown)


jaraco:
Thanks for the contrib. Can you explain why it is that the existing keychain support is insufficient? See [this comment](https://github.com/jaraco/keyring/issues/369#issuecomment-456304754) for one example usage.

a350h:
Targeting a specific keychain at call time allows for much more flexibility in environments where developers have to use multiple keychains in the same script. for instance, environments that use identity tokens in the login chain, but use a separate keychain for API keys for SaaS services. With Stock keyring, one would have to use the hacky method mentioned in your comment, or do some env var magic.

jaraco:
That's reasonable, though I do suspect there's a less invasive way to accomplish this need. I'll make some comments on my concerns with the proposed approach and then suggest an alternative.

README.rst
```diff
@@ -72,6 +72,14 @@ The basic usage of keyring is pretty simple: just call
     >>> keyring.get_password("system", "username")
     'password'
 
+When using keyring with the macOS Keychain, you can pass the full file path of
+a keychain-db to customize which keychain keyring utilizes. Note: the keychain must
+already exist::
+
+    >>> import keyring
+    >>> path = "/Users/foobar/Library/Keychains/Keys.keychain-db"
+    >>> keyring.set_password("system", "username", "password", path)
```
jaraco, line 81: Here, I'd recommend to use a keyword parameter, so it's clearer what `path` is in `set_password`.

keyring/backends/macOS/__init__.py
```diff
@@ -1,5 +1,6 @@
 import platform
 import os
+from os.path import exists
```
jaraco, line 3: I'd avoid this unnecessary import. `os.path.exists` is already available in the namespace.

keyring/backends/macOS/__init__.py
```diff
@@ -17,8 +18,8 @@
 class Keyring(KeyringBackend):
     """macOS Keychain"""
 
-    keychain = os.environ.get('KEYCHAIN_PATH')
-    "Path to keychain file, overriding default"
+    # keychain = os.environ.get('KEYCHAIN_PATH')
+    # "Path to keychain file, overriding default. function param "
```
jaraco, line 22: Commented code is usually a bad smell. Better would be to actually supply the attribute using the default value.

keyring/backends/macOS/__init__.py
```diff
@@ -32,36 +33,54 @@ def priority(cls):
             raise RuntimeError("Security API unavailable")
         return 5
 
-    def set_password(self, service, username, password):
+    def set_password(self, service, username, password, keychain=""):
```
jaraco, line 36: Using `""` is a bad default value... unless the empty string is actually meaningful as a default value.

keyring/backends/macOS/__init__.py
```diff
@@ -32,36 +33,54 @@ def priority(cls):
...
         if username is None:
             username = ''
 
+        if keychain != "":
+            if not exists(keychain):
+                raise KeyringLocked("Can't open specified keychain")
```
jaraco, line 42: I'd not use KeyringLocked here... because the keyring isn't locked. You'll want to create a new exception or find one that's appropriate for the failure.

keyring/backends/macOS/__init__.py
```diff
@@ -32,36 +33,54 @@ def priority(cls):
...
+    def get_password(self, service, username, keychain=""):
         if username is None:
             username = ''
 
+        if keychain != "":
+            if not exists(keychain):
+                raise KeyringLocked("Can't open specified keychain")
+        elif os.environ.get('KEYCHAIN_PATH') != "":
+            keychain = os.environ.get('KEYCHAIN_PATH')
+
```
jaraco, line 62: This code looks like a repeat of something that was done elsewhere. Better would be to implement this logic in one place.

keyring/backends/macOS/__init__.py
```diff
@@ -32,36 +33,54 @@ def priority(cls):
...
+    def set_password(self, service, username, password, keychain=""):
         if username is None:
             username = ''
 
+        if keychain != "":
+            if not exists(keychain):
+                raise KeyringLocked("Can't open specified keychain")
+        elif os.environ.get('KEYCHAIN_PATH') != "":
+            keychain = os.environ.get('KEYCHAIN_PATH')
+
```
jaraco, line 45: There's a whole block of logic here that adds to the complexity of the set_password method. Better than to have it inline would be to have it as a method or function or property that describes its purpose.

keyring/testing/backend_macOS.py
```diff
@@ -0,0 +1,173 @@
+# coding: utf-8
+
+"""
+Common test functionality for backends.
```
jaraco, line 4: This whole module appears to be a copy/paste of the normal tests. That's not good. Now bugs in one module will have to be fixed in the other. More importantly, it's unclear how this test differs from the normal tests. I could figure it out  by inspecting or running a diff on the two tests, but that's a lot of work.

It also means this backend may no longer be adequately tested compared to other backends.

Better is to have exactly one test for each aspect of behavior. In the case of this change, I'd expect to see a new test or class of tests that exercise a new user journey that uses the non-default keychain, probably in test_macOS.

jaraco:
For the reasons above, I'm declining this change, but I'll put something together that may be acceptable.

keyring/testing/backend_macOS.py
```diff
@@ -0,0 +1,173 @@
...
+        monkeypatch.setattr(os, 'environ', env)
+        self.keyring.set_properties_from_env()
+        assert self.keyring.foo_bar == 'fizz buzz'
+
+    def test_keychain_declaration(self):
+        self.keyring.set_password(
+            'service',
+            'user1',
+            'password1',
+            # '/Users/<insert local user here>/Library/Keychains/APIs.keychain-db',
```
jaraco, line 172: Any tests need to capture the desired behavior or be skipped. This test claims to test the keychain declaration, but in fact doesn't do any such thing, but instead has commented code.

PR #576, opened 2022-05-23
Make `importlib_metadata` optional on Python 3.10

### LINKED ISSUE #575: Make importlib_metadata optional on Python 3.10.
> That extra dependency will become optional again in Python 3.10 alpha 7.

Could it be marked as required only for Python < 3.10 in `setup.cfg`?

_Originally posted by @dotlambda in https://github.com/jaraco/keyring/issues/503#issuecomment-1134170726_

tal66:
Hi, I think this will close #575.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,7 +23,7 @@ install_requires =
 	pywin32-ctypes!=0.1.0,!=0.1.1; sys_platform=="win32"
 	SecretStorage>=3.2; sys_platform=="linux"
 	jeepney>=0.4.2; sys_platform=="linux"
-	importlib_metadata >= 3.6
+	importlib_metadata >= 3.6; python_version<"3.10"
 
 [options.packages.find]
 exclude =


mitya57 (changes requested):
You also need to update imports to use the standard library module if it's available, instead of the backport.

I.e. replace this
```python
import importlib_metadata as metadata
```

with something like this
```python
import sys

if sys.version_info >= (3, 10):
    from importlib import metadata
else:
    import importlib_metadata as metadata
```

tal66:
closing this, since I see you gave the same task to another PR with the same change, submitted 2 days after this one.
no need to have us both do this

mitya57:
I don't have a reason to prefer the other pull request over this one, or vice versa. I saw two similar pull requests, and left two similar comments.

The other pull request is still not ready yet — there are CI failures.

tal66:
i'm sure with your guidance he can pull this off. illogical to let us both spend time on this just to reject one. plus i don’t like the encouragement to submit a pr that someone else already started – i’d never do that even when I know better. cheers

PR #577, opened 2022-05-25
Changed importlib_metadata to have requirement for < python 3.10

probro27:
Trying to fix the issue #575.
Hope it works!!

PR #578, opened 2022-06-02
Use native importlib.metadata when possible.

kousu:
The importlib_metadata backport is only needed for python 3.7 now, since python 3.6 is at End-of-Life, but keyring makes it, and its dependency zipp get downloaded and installed.

### DIFF
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -2,6 +2,7 @@
 Keyring implementation support
 """
 
+import sys
 import os
 import abc
 import logging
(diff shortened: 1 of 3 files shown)


mitya57:
Hi!

This is third pull request for `importlib.metadata` change. We have already had #576 and #577.

#577 is approved, I hope we will merge it soon. Note that we need to use `importlib.metadata` only for Python ≥ 3.10, not ≥ 3.8, because 3.8 and 3.9 don't support `group` argument (see the failed tests).

So closing this in favor of #577.

PR #582, opened 2022-06-28
Let devpi fall back to other options if keyring fails

takluyver:
keyring provides a plugin to supply the password to devpi-client. However, if keyring fails - e.g. it's installed but no backend is working, or unlocking the keyring fails - then `devpi login` becomes impossible, because the keyring exception is uncaught.

This catches any KeyringError, and returns None instead, which signals to devpi that it should try another way to get the password ([docs](https://devpi.net/docs/devpi/devpi/stable/+d/devguide/hooks.html#devpi-client-plugin-hooks-experimental)).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+v23.7.0
+-------
+
+* #582: Suppress KeyringErrors for devpi client.
+
 v23.6.0
 -------
 
--- a/keyring/devpi_client.py
+++ b/keyring/devpi_client.py
@@ -1,6 +1,7 @@
 from pluggy import HookimplMarker
 
 import keyring
+from keyring.errors import KeyringError
 
 
 hookimpl = HookimplMarker("devpiclient")
@@ -8,4 +9,7 @@ hookimpl = HookimplMarker("devpiclient")
 
 @hookimpl()
 def devpiclient_get_password(url, username):
-    return keyring.get_password(url, username)
+    try:
+        return keyring.get_password(url, username)
+    except KeyringError:
+        return None
--- a/keyring/devpi_client.py
+++ b/keyring/devpi_client.py
@@ -1,5 +1,6 @@
+import contextlib
+
 from pluggy import HookimplMarker
-from jaraco.context import suppress
 
 import keyring
 from keyring.errors import KeyringError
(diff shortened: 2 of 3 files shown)


jaraco:
LGTM. I'd prefer to use the suppress decorator, but I'll use your change as inspiration.

jaraco:
In this latest patch, I add the use of the jaraco.context.suppress as a simple, elegant decorator. Unfortunately, because that behavior isn't in the stdlib, it adds a dependency, which I just can't justify for an entry point like this, so I'll find another way.

eachimei:
Hello,
There's an issue with this recent change and how `pluggy` (which is used by `devpi-client`) checks the hook-impl.
The short story is: `devpi-client` uses `pluggy` -->`pluggy` uses `inspect.getfullargspec(func)` to get the implementation arguments --> `inspect.getfullargspec(func)` does not work well once the function is decorated with `contextlib.ContextDecorator` --> as a result, pluggy "thinks" the hookimpl doesn't expect any arguments and hence it does not forward the provided arguments to the hookimpl function.
As a result:

```
$ python -m devpi login <...>
  File "C:\Python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "...\site-packages\devpi\__main__.py", line 3, in <module>
    main.main()
  File "...\site-packages\devpi\main.py", line 36, in main
    return method(hub, hub.args)
  File "...\site-packages\devpi\login.py", line 13, in main
    password = hub.hook.devpiclient_get_password(
  File "...\site-packages\pluggy\_hooks.py", line 265, in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
  File "...\site-packages\pluggy\_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  File "...\site-packages\pluggy\_callers.py", line 60, in _multicall
    return outcome.get_result()
  File "...\site-packages\pluggy\_result.py", line 60, in get_result
    raise ex[1].with_traceback(ex[2])
  File "...\site-packages\pluggy\_callers.py", line 39, in _multicall
    res = hook_impl.function(*args)
  File "C:\Python38\lib\contextlib.py", line 75, in inner
    return func(*args, **kwds)
TypeError: devpiclient_get_password() missing 2 required positional arguments: 'url' and 'username'
```

I tested with Python 3.8 and 3.9 (same behavior).

Looks like a backwards-compatibility decision in `inspect.getfullargspec` regarding `follow_wrapper_chains=False`.

A suggestion for `pluggy` might be: use `inspect.signature(func)` instead of `getfullargspec` since it has the ability to follow wrapper chains (and that's actually the default behavior): https://docs.python.org/3/library/inspect.html#inspect.signature

@jaraco

PR #586, opened 2022-08-05
Allow scheme to be selectable in libsecret and SecretService backends

### LINKED ISSUE #448: Using python-keyring with KeePassXC
**Describe the bug**

The format of data in a KeePassXC database is different from what the `SecretService` backend expects, which means there's no way to get data created natively in KeePassXC.

**To Reproduce**

1. Download KeePassXC (2.6.0 has an appimage version, just download and run)
1. Create a new database for testing
1. Enable Secret Service Integration, expose your database
1. Create a new entry, title `foo`, username `bar`, password `baz`
1. Create a new entry from python-keyring in the database, but this time using python-keyring:

    ```python
    import keyring as k
    k.set_keyring(k.backends.SecretService.Keyring())
    k.set_password('kfoo', 'kbar', 'kbaz')
    ```

1. Observe the different entry layouts. (The attributes set by python-keyring can be found by right-clicking the entry, then "Properties", then "Advanced" on the left.)
1. Note that there's no way to get at a "natively created" KeePassXC entry using the Secret Service backend.

**Expected behavior**

- `keyring.get_password()` works correctly through KeePassXC's Secret Service integration.
- A way to get entries created in KeePassXC using python-keyring.

**Environment**

- OS: Fedora 32

```
$ pip list | grep keyring
keyring 21.2.0
$ keyring --list-backends
keyring.backends.chainer.ChainerBackend (priority: 10)
keyring.backends.kwallet.DBusKeyring (priority: 5.1)
keyring.backends.SecretService.Keyring (priority: 5)
```

**Additional context**

What I would *really* like to do is to use the [Ansible `keyring` lookup plugin](https://docs.ansible.com/ansible/latest/plugins/lookup/keyring.html) to find credentials stored in KeePassXC, but that won't be possible unless python-keyring understands the format provided by KeePassXC, I believe.

This is how secret-tool formats the KeePassXC generated entry:

```ini
[/org/freedesktop/secrets/collection/runiq/6318916706d54321a7ffd0b0a744c05a]
label = foo
secret = baz
created = 2020-07-15 17:18:11
modified = 2020-07-15 17:18:19
attribute.Title = foo
attribute.URL =
attribute.UserName = bar
attribute.Path = /foo
attribute.Uuid = 6318916706d54321a7ffd0b0a744c05a
attribute.Notes =
```

And this is the python-keyring generated entry:

```ini
[/org/freedesktop/secrets/collection/runiq/e50ead9611834503acd02bedd3791ce4]
label = Password for 'kbar' on 'kfoo'
secret = kbaz
created = 2020-07-15 18:00:25
modified = 2020-07-15 18:00:25
attribute.URL =
attribute.Title = Password for 'kbar' on 'kfoo'
attribute.UserName = runiq
attribute.Notes =
attribute.application = Python keyring library
attribute.Uuid = e50ead9611834503acd02bedd3791ce4
attribute.username = kbar
attribute.Path = /Password for 'kbar' on 'kfoo'
attribute.service = kfoo
```

jaraco:
- Add support for a scheme property in SecretService backend. Set to 'KeypassXC' to use the username/system convention from KeypassXC, either with KEYRING_PROPERTY_SCHEME or keyring.with_properties(scheme='KeypassXC'). Fixes #448.
- Add test for with_properties helper.
- Re-use with_properties in with_keychain.
- Deprecate macOS.Keyring.with_keychain, superseded by with_properties.
- Extract SchemeSelectable mixin
- Allow _query to include other keys
- Add SelectableScheme to libsecret.
- Add tests for SchemeSelectable.

### DIFF
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -224,6 +224,17 @@ class SchemeSelectable:
     """
     Allow a backend to select different "schemes" for the
     username and service.
+
+    >>> backend = SchemeSelectable()
+    >>> backend._query('contoso', 'alice')
+    {'username': 'alice', 'service': 'contoso'}
+    >>> backend._query('contoso')
+    {'service': 'contoso'}
+    >>> backend.scheme = 'KeypassXC'
+    >>> backend._query('contoso', 'alice')
+    {'UserName': 'alice', 'Title': 'contoso'}
+    >>> backend._query('contoso', 'alice', foo='bar')
+    {'UserName': 'alice', 'Title': 'contoso', 'foo': 'bar'}
     """
 
     scheme = 'default'
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -218,3 +218,29 @@ def get_all_keyring():
     viable_classes = KeyringBackend.get_viable_backends()
     rings = util.suppress_exceptions(viable_classes, exceptions=TypeError)
     return list(rings)
+
+
+class SchemeSelectable:
+    """
+    Allow a backend to select different "schemes" for the
+    username and service.
+    """
+
+    scheme = 'default'
+    schemes = dict(
+        default=dict(username='username', service='service'),
+        KeypassXC=dict(username='UserName', service='Title'),
+    )
+
+    def _query(self, service, username):
+        scheme = self.schemes[self.scheme]
+        return (
+            {
+                scheme['username']: username,
+                scheme['service']: service,
+            }
+            if username
+            else {
+                scheme['service']: service,
+            }
+        )
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -232,9 +232,9 @@ class SchemeSelectable:
         KeypassXC=dict(username='UserName', service='Title'),
     )
 
-    def _query(self, service, username):
+    def _query(self, service, username=None, **base):
         scheme = self.schemes[self.scheme]
-        return (
+        return dict(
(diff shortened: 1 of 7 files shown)


mitya57 (approved):

keyring/backend.py
```diff
@@ -212,3 +218,41 @@ def get_all_keyring():
...
+    >>> backend._query('contoso', 'alice')
+    {'UserName': 'alice', 'Title': 'contoso'}
+    >>> backend._query('contoso', 'alice', foo='bar')
+    {'UserName': 'alice', 'Title': 'contoso', 'foo': 'bar'}
+    """
+
+    scheme = 'default'
+    schemes = dict(
+        default=dict(username='username', service='service'),
+        KeypassXC=dict(username='UserName', service='Title'),
```
mitya57: It should be `KeePassXC` (with `e` instead of `y`):

https://keepassxc.org/

Please update it in all other places too.

mitya57 (changes requested):

keyring/backend.py
jaraco: Fixed in d35fc44.

jaraco:
Released as 23.8.0.

mitya57:
I am sorry for not testing it earlier, but libsecret backend does not work :(

Submitted #587.

PR #587, opened 2022-08-07
Fix regression which made libsecret backend unusable in v23.8.0

mitya57:
libsecret/secret-schema.h has this enum:

```c
typedef enum {
    SECRET_SCHEMA_ATTRIBUTE_STRING = 0,
    SECRET_SCHEMA_ATTRIBUTE_INTEGER = 1,
    SECRET_SCHEMA_ATTRIBUTE_BOOLEAN = 2,
} SecretSchemaAttributeType;
```

Because of this, `bool(Secret.SchemaAttributeType.STRING)` evaluates to False. So when we do this, _query code ignores the second argument:

https://github.com/jaraco/keyring/blob/0ed196d309f657b69c0f9d76d7c9e5020b7e0b54/keyring/backends/libsecret.py#L40-L44

And it causes any use of libsecret backend to fail with this error:

    >>> import keyring.backends.libsecret
    >>> k = keyring.backends.libsecret.Keyring()
    >>> k.set_password('foo', 'bar', 'baz')

    (process:26311): libsecret-CRITICAL **: 19:22:53.299: secret_password_storev_sync: invalid username attribute for org.freedesktop.Secret.Generic schema
      File "<stdin>", line 1, in <module>
      File "/home/dmitry/upstream/keyring/keyring/backends/libsecret.py", line 98, in set_password
        raise PasswordSetError("Failed to store password!")
    keyring.errors.PasswordSetError: Failed to store password!

### DIFF
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -250,7 +250,7 @@ class SchemeSelectable:
                 scheme['username']: username,
                 scheme['service']: service,
             }
-            if username
+            if username is not None
             else {
                 scheme['service']: service,
             },
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+v23.8.1
+-------
+
+* #587: Fix regression in ``libsecret``.
+
 v23.8.0
 -------
 


keyring/backend.py
```diff
@@ -250,7 +250,7 @@ def _query(self, service, username=None, **base):
                 scheme['username']: username,
                 scheme['service']: service,
             }
-            if username
+            if username is not None
```
jaraco, line 253: It's possible some users are relying on `username = ''` as a non-value for username. We can address that separately if needed.

jaraco:
Thanks for the patch. I'd like to develop a regression test for the missed expectation. I'll do that and then merge it.

jaraco:
I realize now we don't have libsecret tested anywhere, so I'm going to pull the test.

jaraco:
Released as v23.8.1.

PR #588, opened 2022-08-08
Use jaraco.classes for properties

jaraco:
Today I was reviewing the implementation of `NonDataProperty` and `ClassProperty` only to find that there were stale versions of the implementations in `keyring.util.properties` that represent old versions of the original implementations in `jaraco.classes`.

I'm tempted to simply depend on that code. Because it introduces a new dependency, I'm somewhat reluctant to accept the change. I'm putting it here for feedback and consideration.


Darsstar:
Maybe reexport them for the time being for backward compatibility? (I'm tempted to replace "Maybe" with "At least", but I feel that makes me sounds too entitled.)

jaraco:
> Maybe reexport them for the time being for backward compatibility? (I'm tempted to replace "Maybe" with "At least", but I feel that makes me sounds too entitled.)

I can do that - make it an optional dependency for the time being.

Thaodan:
Isn't that kinda much to add another dependency just for these two methods?

jaraco:
Yes and no.

1. In an ideal world, dependencies are free. And for most users, that's also true (`pip install keyring` just works in all cases). Of course, there are practical, non-ideal considerations.
2. There are several costs to limiting dependencies. It increases friction when creating new functionality. It encourages vendoring (maintaining a local fork) of the functionality, which has its own maintenance costs and risks. It discourages best practices around dependency management.
3. Not using the dependency masks the true dependency. It hides the fact that this functionality is useful across multiple projects, making it nearly impossible to determine if this shared need would be valuable to be exposed elsewhere (stdlib, for example).
4. This case is a perfect case for including the dependency. There was an old, stale implementation that was copied from many years ago. Since then, the implementation has improved and evolved. In the meantime, other libraries developed a dependency on this vendored fork of the code.

Principles I roughly follow for dependencies:

- If a small (~10 lines or less) piece of code is useful, it's okay to copy it, but still link to the source from which it was copied.
- Consider using a dependency when the code is larger, there is more than one function, or the behavior is likely to evolve.
- Take extra precautions when adopting the first dependency, as the distance from 0 to 1 is much greater than adding one more.

I think implicit in your question is that adding a dependency is a cost that should be taken seriously. Can you articulate or link to a resource that explains those costs?

yan12125:
```Python
try:
    from jaraco.compat import properties  # pragma: no-cover
except ImportError:
    from . import _properties_compat as properties  # pragma: no-cover
```
I think `from jaraco.compat import properties` should be `from jaraco.classes import properties`?

PR #598, opened 2022-09-17
Apply workaround for decoratated function unrecognized

### LINKED ISSUE #596: devpi-client hook broken
The [``@suppress(KeyringError)``](https://github.com/jaraco/keyring/commit/314c8d36fa06090938f95d2672316471b15a6e65#diff-b0e6ff13ff76becf8aee3446689be4103ae1db3b2ba31748c7cb7b934affc033R17) line breaks the devpi-client hook. It looks like the argument spec is changed by it which causes pluggy to not pass in the arguments correctly. It happens when trying to ``devpi push`` a release to pypi and the password is not set in ``~pypirc`` which causes devpi-client to try the hook.

jaraco:
- Add test to ensure varnames are calculated correctly. Ref #596.
- Add workaround for pytest-dev/pluggy#358. Fixes #596.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+v23.9.3
+-------
+
+* #596: Add workaround for devpi_client hook with wrapped implementation.
+
 v23.9.2
 -------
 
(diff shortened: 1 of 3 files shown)


jaraco:
Although it would have been straightforward to remove the 'suppress' decorator and inline it with the hook, that would have masked the issue upstream. This approach takes another tack, which is to restore the necessary signature after the decorator, such that the restoration can be removed once the upstream bug is fixed.

PR #603, opened 2022-10-31
libsecret: skip if there is no secret service running

### LINKED ISSUE #600: Crash with gi.repository.GLib.GError: g-dbus-error-quark: The name org.freedesktop.secrets was not provided by any .service files
**Describe the bug**
It crashes

**To Reproduce**
1. Run `keyring set system test`
2. Observe the exception
```
Password for 'test' in 'system': 
  File "/usr/bin/keyring", line 8, in <module>
    sys.exit(main())
  File "/usr/lib/python3.10/site-packages/keyring/cli.py", line 132, in main
    return cli.run(argv)
  File "/usr/lib/python3.10/site-packages/keyring/cli.py", line 68, in run
    return method()
  File "/usr/lib/python3.10/site-packages/keyring/cli.py", line 85, in do_set
    set_password(self.service, self.username, password)
  File "/usr/lib/python3.10/site-packages/keyring/core.py", line 60, in set_password
    get_keyring().set_password(service_name, username, password)
  File "/usr/lib/python3.10/site-packages/keyring/backends/libsecret.py", line 85, in set_password
    stored = Secret.password_store_sync(
gi.repository.GLib.GError: g-dbus-error-quark: The name org.freedesktop.secrets was not provided by any .service files (2)
```

**Expected behavior**
Not an error

**Environment**
- Arch Linux with python-keyring 23.9.3-1

```
$ pip list | grep keyring
keyring              23.9.3

$ keyring --list-backends
keyring.backends.libsecret.Keyring (priority: 4.8)
keyring.backends.chainer.ChainerBackend (priority: -1)
```

**Additional context**
At this is now a dependency of poetry, I cannot run poetry anymore. I can workaround this using:
```bash
export PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring
```

lazka:
In case libsecret and its Python bindings are available the libsecret backend will get selected. This doesn't mean though that the system service it connects to is available.

This leads to the backend being used and then failing, for example in get_credential() with:
g-dbus-error-quark: The name org.freedesktop.secrets is unknown (2)

To avoid this ask libsecret to connect to the system service first before considering to use it.

Fixes #600

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+v23.11.0
+--------
+
+* #603: In ``libsecret``, check that the service is available before
+  declaring viability.
+
 v23.10.0
 --------
 
--- a/keyring/backends/libsecret.py
+++ b/keyring/backends/libsecret.py
@@ -7,7 +7,6 @@ from ..credentials import SimpleCredential
 from ..errors import (
     PasswordDeleteError,
     PasswordSetError,
-    ExceptionRaisedContext,
     KeyringLocked,
 )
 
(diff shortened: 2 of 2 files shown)


lazka:
I noticed priority gets called quite a lot (~8 times), so the priortiy/viable split that the code hints at seems like a good idea. In this case everything after the first call is quite fast with libsecret, so this shouldn't be a blocker. Just something I've noticed.

lazka:
(oops, small conflict as I was trying to force push the black CI error fix at the same time, should be fixed again)

lazka:
Thanks!

Issue #573, opened 2022-05-10
Setting a blank password causes subsequent 'set' calls to fail


AlexanderVatov:
**Describe the bug**
If a password is set to '', any subsequent attempts to set it to another value result in the following error:

```
Traceback (most recent call last):
  File "/Library/Python/3.8/site-packages/keyring/backends/macOS/__init__.py", line 40, in set_password
    api.set_generic_password(self.keychain, service, username, password)
  File "/Library/Python/3.8/site-packages/keyring/backends/macOS/api.py", line 161, in set_generic_password
    Error.raise_for_status(status)
  File "/Library/Python/3.8/site-packages/keyring/backends/macOS/api.py", line 114, in raise_for_status
    raise cls(status, "Unknown Error")
keyring.backends.macOS.api.Error: (-25299, 'Unknown Error')
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/Library/Python/3.8/site-packages/keyring/core.py", line 60, in set_password
    get_keyring().set_password(service_name, username, password)
  File "/Library/Python/3.8/site-packages/keyring/backends/macOS/__init__.py", line 44, in set_password
    raise PasswordSetError("Can't store password on keychain: " "{}".format(e))
keyring.errors.PasswordSetError: Can't store password on keychain: (-25299, 'Unknown Error')
```

**To Reproduce**
```python
import keyring
keyring.set_password('test-keyring-123', 'test', 'a') # Succeeds
keyring.set_password('test-keyring-123', 'test', 'b') # Succeeds
keyring.set_password('test-keyring-123', 'test', '')  # Succeeds
keyring.set_password('test-keyring-123', 'test', 'a') # Fails
```

**Expected behavior**
One would expect set_password not to be affected by previous calls of set_password.

**Environment**

Python 3.8.9 on macOS

``` 
$ pip list | grep keyring
keyring                 23.5.0
$ keyring --list-backends
keyring.backends.chainer.ChainerBackend (priority: -1)
keyring.backends.macOS.Keyring (priority: 5)
keyring.backends.fail.Keyring (priority: 0)
```

**Additional context**
Getting the blank password works fine. Deleting it and then setting it to any non-blank password also works as a workaround.

IbrahimAH:
[...]

PR #614, opened 2022-12-15
bugfix/573

IbrahimAH:
per comment in #573, if user stores an empty string, it is returned to the if statement as `''`. in python, `''` evaluates as false, resulting in it not deleting the secret before trying to set a new one. this results in an error

### DIFF
--- a/keyring/backends/macOS/api.py
+++ b/keyring/backends/macOS/api.py
@@ -147,7 +147,7 @@ def find_generic_password(kc_name, service, username, not_found_ok=False):
 
 
 def set_generic_password(name, service, username, password):
-    if find_generic_password(name, service, username, not_found_ok=True):
+    if find_generic_password(name, service, username, not_found_ok=True) is not None:
         delete_generic_password(name, service, username)
 
     q = create_query(


IbrahimAH:
i need some pointers here on why its failing ubuntu tests if the only changes were to the macos backend!

jaraco:
Test failure was due to issues in the packaging infrastructure relating to new releases of tox, fixed in main.

keyring/backends/macOS/api.py
```diff
@@ -147,7 +147,7 @@ def find_generic_password(kc_name, service, username, not_found_ok=False):
 
 
 def set_generic_password(name, service, username, password):
-    if find_generic_password(name, service, username, not_found_ok=True):
+    if find_generic_password(name, service, username, not_found_ok=True) is not None:
         delete_generic_password(name, service, username)
```
jaraco, line 151: Even better, perhaps one should not pass `not_found_ok` and trap the `NotFound` error. Even better than that, invoke the deletion and suppress the `NotFound`. I'll do that separately. Thanks for raising this to my attention.

PR #617, opened 2023-01-16
Fix a typo of README.rst

Freed-Wu:


mitya57:
It's not a typo, it's someone tried to use Markdown syntax in a reStructuredText file :)

Merged, thanks!

PR #625, opened 2023-03-27
Add more type annotations

lovetox:
Hi,

I tried to add more type annotations

- The docs say priority must be defined and is a property and must return float or int, but in the base class it was a method returning None. I tried to make this more consistent across the backends to make type checkers happy.

### DIFF
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -1,6 +1,7 @@
 """
 Keyring implementation support
 """
+from __future__ import annotations
 
 import os
 import abc
@@ -8,7 +9,7 @@ import logging
 import operator
 import copy
 
-from typing import Optional
+import typing
 
 from .py312compat import metadata
 from . import credentials, errors, util
@@ -18,7 +19,7 @@ log = logging.getLogger(__name__)
 
 
 by_priority = operator.attrgetter('priority')
-_limit = None
+_limit: typing.Optional[typing.Callable[[KeyringBackend], bool]] = None
 
 
 class KeyringBackendMeta(abc.ABCMeta):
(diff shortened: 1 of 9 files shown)


keyring/backend.py
```diff
@@ -45,8 +45,8 @@ class KeyringBackend(metaclass=KeyringBackendMeta):
     def __init__(self):
         self.set_properties_from_env()
 
-    # @abc.abstractproperty
```
jaraco, line 47: The intention here is to eventually make this method an abstractproperty so that a subclass would fails to construct if it doesn't define the property (failing faster).

It's my understanding that `NotImplementedError` is an older paradigm that preceded use of `abstractproperty`.

Wouldn't it be better to implement this as an abstract property? Does the typing system understand that an abstract property doesn't need an implementation?

jaraco:
I've rebased the changes on top of 687d2c5 (fixing the emergent black failure).

keyring/backend.py
```diff
@@ -144,14 +144,11 @@ def get_credential(
...
-            return sep and (name.lower(), value)
-
-        props: filter[typing.Tuple[str, str]] = filter(None, map(parse, os.environ.items()))
-        for name, value in props:
-            setattr(self, name, value)
+        for key, value in os.environ.items():
+            _, sep, name = key.partition('KEYRING_PROPERTY_')
+            if not sep:
+                continue
+            setattr(self, name.lower(), value)
```
jaraco: I'd prefer to use a functional paradigm. I do recognize how the imperative approach is simpler, especially when type annotations are present. Still, it's important for me when adding typing hints not to allow the addition of hints to change the semantic meaning of the implementation, so I've backed out this commit.

keyring/backend.py
jaraco: (eventually because it's liable to cause a backward incompatibility)

lovetox: But a abstractproperty is not a classproperty.

So if you want to have priority defined on class level there is no abstract concept for it to my knowledge in python.

keyring/backend.py
lovetox: fine by me, but what do you mean by "backed out this commit", nothing is backed out, its still here.

Do you mean i should remove it?

lovetox: Ah sorry see it now, you reverted it, i would find it nicer if i simply remove the commit if its fine by you.

jaraco: Yes, feel free to rewrite the history without those changes.

keyring/backend.py
jaraco: Ah. Good point and thanks for the reminder. In that case, then yes, I agree wit the change.

lovetox:
I remove the commit, so everything is ready from my side

PR #638, opened 2023-06-23
Fix reading config file

jteichroeb-oanda:
_ensure_path never returns a value causing the config file to never be read. Fix the existing test to ensure that the file it's looking for can't exist, add a new test that ensure the file does exist

### DIFF
--- a/keyring/core.py
+++ b/keyring/core.py
@@ -153,6 +153,7 @@ def _config_path():
 def _ensure_path(path):
     if not path.exists():
         raise FileNotFoundError(path)
+    return path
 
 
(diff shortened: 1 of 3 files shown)


zoni (approved):
Can confirm `24.1.0` (latest on PyPI at time of writing) suffers from this issue and that this patch will resolve the problem.

Would go so far as to recommend cutting a `24.1.1` patch release with this fix in it, as this library currently throws an exception for any user who has a config file present.

jaraco:
Releasing as v24.1.1 (with Python 3.7 support) and v24.2.0.

PR #643, opened 2023-07-28
Completion improvements

scop:
See individual commits for details.

### DIFF
--- a/keyring/completion.py
+++ b/keyring/completion.py
@@ -39,8 +39,8 @@ def install_completion(parser):
         zsh=files(__package__).joinpath('backend_complete.zsh').read_text(),
     )
     shtab.add_argument_to(parser, preamble=preamble)
-    get_action(parser, '--keyring-path').completion = shtab.DIR
-    get_action(parser, '--keyring-backend').completion = dict(zsh='backend_complete')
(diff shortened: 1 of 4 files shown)


keyring/completion.py
```diff
@@ -36,11 +36,15 @@ def get_action(parser, option):
...
     preamble = dict(
+        bash=files(__package__).joinpath('backend_complete.bash').read_text(),
         zsh=files(__package__).joinpath('backend_complete.zsh').read_text(),
     )
     shtab.add_argument_to(parser, preamble=preamble)
-    get_action(parser, '--keyring-path').completion = shtab.DIR
-    get_action(parser, '--keyring-backend').completion = dict(zsh='backend_complete')
+    get_action(parser, '--keyring-path').complete = shtab.DIR
+    get_action(parser, '--keyring-backend').complete = dict(
+        bash='_keyring_backends',
```
scop, line 45: I chose not to follow the zsh function name precedent for bash, as `backend_complete` is a bit generic, and _could_ clash with a user function or command. Prefixing with the underscore is a common convention to try and avoid that.

zsh could possibly follow suit, but I'll leave that for someone else to address :)

jaraco (approved):

PR #655, opened 2023-10-10
Find password even with no username supplied in OSX backend

maarten0912:
Find a keyring password even if no username is supplied in the OS_X backend

### DIFF
--- a/keyring/backends/macOS/__init__.py
+++ b/keyring/backends/macOS/__init__.py
@@ -57,9 +57,6 @@ class Keyring(KeyringBackend):
 
     @warn_keychain
     def get_password(self, service, username):
-        if username is None:
-            username = ''
-
         try:
             return api.find_generic_password(self.keychain, service, username)
         except api.NotFound:
--- a/keyring/backends/macOS/api.py
+++ b/keyring/backends/macOS/api.py
@@ -128,11 +128,13 @@ class SecAuthFailure(Error):
 
 
 def find_generic_password(kc_name, service, username, not_found_ok=False):
+    username_kwarg = {'kSecAttrAccount': username} if username is not None else {}
+        
     q = create_query(
         kSecClass=k_('kSecClassGenericPassword'),
         kSecMatchLimit=k_('kSecMatchLimitOne'),
         kSecAttrService=service,
-        kSecAttrAccount=username,
+        **username_kwarg,
         kSecReturnData=create_cfbool(True),
     )
 


jaraco:
Can you tell me more about the motivations for making this change? Under what conditions is a null versus empty username relevant for macOS? What compatibility considerations should we make for users relying on the current behavior (None -> "")?

jaraco:
Any thoughts on my questions?

maarten0912:
> Any thoughts on my questions?

The motivation for the change is because I wanted to find a keyring based only on its name (kSecAttrService). In my use case, the account of the keyring (kSecAttrAccount, but called username above) was not known and not important. The API will give the first result in case there are multiple keyring with the specified name, which was what I wanted in my case.

I would say that the library should work for querying empty usernames (account name is "") of keyrings and also for querying without a username (account name is None). Unfortunately, I do not see what we can do for users relying on the current behaviour.

jaraco:
I don't have a good understanding of the differences between the following scenarios:

- `kSetAttrAccount` = ''
- `kSetAttrAccount` = None
- no `kSetAttrAccount`

I'm guessing `kSetAttrAccount` of None is invalid.

More importantly, we'll probably want to align with other backends to provide an expectation at the keyring level of what should happen for empty or null usernames.

Since there do not appear to be any tests protecting the existing behavior (`None` → &lt;empty string>), we should at least trace that code and determine when that logic was added to see what the motivation might have been for it.

jaraco (changes requested):
Interesting - it appears as if that logic was added when the API was [migrated from C to ctypes](https://github.com/jaraco/keyring/commit/978f9b187d636bd49b16c5a9ff1b44ba9130eb5f). The C logic also [translated None to the empty string](https://github.com/jaraco/keyring/commit/978f9b187d636bd49b16c5a9ff1b44ba9130eb5f#diff-2413567858269fc62b79e927f0df01e9b9ea0ef6a084e7b52ea1cc25eba06ebdL31-L33).

This all makes me think the logic is more likely incidental than necessary... and support for None was minimal. Indeed, there are no tests for it and [the type spec indicates a string is required](https://github.com/jaraco/keyring/blob/3727268f0de9d5ab56d94e2cff0a794153769c18/keyring/backend.py#L100).

A couple of requests: 

- [ ] Can you add a test or two to cover the new expectation?
- [ ] Can you verify what happens when `None` is passed explicitly for kSecAttrAccount?
- [ ] Can you unify the logic across all methods (get/set/del/...)?

jaraco:
I have some news regarding this issue. In #668 and #687, I'm exploring completely deprecating support for empty usernames across all keyrings to provide consistency. The guidance in that change is that all clients of keyring should always pass a non-empty username, even if it's just some static value. Some backends, like Windows, don't behave properly if the username field is empty, and I want to try to avoid inconsistencies (so an application doesn't test on Linux or Mac and then find that users on Windows get a bad experience).

If support for empty usernames is valuable to you, can you comment in #668 as to why? Thanks.

Issue #656, opened 2023-10-18
Feature request: CLI option to output password without newline?


wcooley:
I would like to request an option to output the password without the trailing newline. I have found a number of cases where the newline ends up being included as part of the password, such as the password file that the OpenLDAP clients can use (`-y <file>`) and when piping into `pbcopy` on macos (to paste into a GUI dialog box).

I would even like to go so far as to request not only a parameter for the command line but to suggest that `os.isatty(sys.stdout.fileno())`(or something similar) be used to exclude the newline when stdout isn't the terminal. But this would be a breaking change, so such a change might not be feasible.

(I know that I can easily strip the newline in various ways, like `|tr -d '\n'`, but that is much less convenient.)

PR #657, opened 2023-10-19
Add argument to not output newline with "get"

wcooley:
This PR addresses #656 which I opened earlier today. It doesn't do anything with completion support; I have to confess that I do not know how that works.

While this could be more sophisticated (such as supporting a `--newline` for symmetry, to undo the effect of a prior `--no-newline`), this is enough to get started and for my own personal needs.

### DIFF
--- a/keyring/cli.py
+++ b/keyring/cli.py
@@ -42,7 +42,8 @@ class CommandLineTool:
             action="store_false",
             default=True,
             dest="include_newline",
-            help="Exclude newline from 'get' output")
+            help="Exclude newline from 'get' output",
+        )
         self.parser._operations = ["get", "set", "del", "diagnose"]
         self.parser.add_argument(
             'operation',
--- a/keyring/cli.py
+++ b/keyring/cli.py
@@ -37,6 +37,12 @@ class CommandLineTool:
         self.parser.add_argument(
             "--disable", action="store_true", help="Disable keyring and exit"
         )
+        self.parser.add_argument(
+            "--no-newline",
+            action="store_false",
+            default=True,
+            dest="include_newline",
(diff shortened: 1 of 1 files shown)


mitya57 (approved):

jaraco:
Thanks for the contrib. I have some reservations, but nothing that can't be overcome.

My reservations stem from, "why hasn't this issue been solved elsewhere?" In other words, it seems to me that keyring isn't the first application to output text in a CLI for use in a pipe. Is there an established convention for handling these sorts of situations? Is there an existing library that implements these conventions (such that they don't have to be implemented inline with keyring's unique logic)? I suspect the answer is "no, every application just does its own thing and it's the responsibility of each application to adapt to its users' needs as the arise," but I'd be delighted if there were some guidance out there.

My other reservation is with the use of a boolean parameter. I try to avoid the use of those. I'd prefer something like "--format" or "--template", which specifies the format of the output, and maybe default to "{password}\n" but allow it to be overridden by "--format {password}". I'm less sure about this concern, and it depends a lot on what background we might discover in the first concern.

Let me know what you find / think.

jaraco:
Another consideration - maybe keyring should detect if the output is being piped, and in that case, default to a format without the newline.

jaraco:
What do you think about re-using the [is tty logic](https://github.com/jaraco/keyring/blob/3727268f0de9d5ab56d94e2cff0a794153769c18/keyring/cli.py#L124-L132) to infer whether the caller would prefer a newline?

jaraco:
Since this issue is languishing, I'm going to close it for now. Feel free to revive the conversation and we can re-open it.

PR #660, opened 2023-11-13
Remove useless shebang from cli.py

mtelka:
This is to solve the following issue:
```
$ keyring/cli.py
  File "/opt/src/keyring/keyring/cli.py", line 8, in <module>
    from . import core
ImportError: attempted relative import with no known parent package
$
```


jaraco:
Thanks for this! It's likely a remnant of a time when it was a standalone script or when Python modules were identified by a generic shebang.

PR #662, opened 2023-11-19
Add: password entry verification to prevent password mistype

abastola0:
Added a password retype mechanism that helps user verify the password in case the password is mistyped. The number of trials for incorrect passoword attempts is set to 3 for now.

### DIFF
--- a/keyring/cli.py
+++ b/keyring/cli.py
@@ -89,20 +89,24 @@ class CommandLineTool:
         trials = 3
         for i in range(trials):
             password = self.input_password(
-                f"Password for '{self.username}' in '{self.service}': ")
-    
+                f"password for '{self.username}' in '{self.service}': "
+            )
+
             reentered_password = self.input_password(
-                f"Re-enter password for '{self.username}' in '{self.service}': ")
-    
+                f"re-enter password for '{self.username}' in '{self.service}': "
+            )
+
             if password == reentered_password:
-                # If passwords match, set the password and break the loop
+                # if passwords match, set the password and break the loop
                 set_password(self.service, self.username, password)
                 break
             else:
-                print('Password verification failed! Try again!\n')
-    
-        else:  # This else is associated with the for-loop, not the if-statement
-            sys.stderr.write(f"Password verification failed for {trials} times. Aborting!!\n")
+                print('password verification failed! try again!\n')
+
+        else:  # this else is associated with the for-loop, not the if-statement
+            sys.stderr.write(
+                f"password verification failed for {trials} times. aborting!!\n"
+            )
             sys.exit(1)
 
     def do_del(self):
--- a/keyring/cli.py
+++ b/keyring/cli.py
@@ -86,10 +86,26 @@ class CommandLineTool:
         print(password)
 
     def do_set(self):
-        password = self.input_password(
-            f"Password for '{self.username}' in '{self.service}': "
-        )
-        set_password(self.service, self.username, password)
+        trials = 3
+        for i in range(trials):
+            password = self.input_password(
+                f"Password for '{self.username}' in '{self.service}': ")
+    
+            reEnteredPassword = self.input_password(
+                f"Re-enter password for '{self.username}' in '{self.service}': ")
+    
+            if password == reEnteredPassword:
+                # If passwords match, set the password and break the loop
+                set_password(self.service, self.username, password)
+                break
+            else:
+                print('Password verification failed! Try again!')
+    
+        else:  # This else is associated with the for-loop, not the if-statement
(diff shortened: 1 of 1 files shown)


keyring/cli.py
```diff
@@ -86,10 +86,26 @@ def do_get(self):
...
+            if password == reEnteredPassword:
+                # If passwords match, set the password and break the loop
+                set_password(self.service, self.username, password)
+                break
+            else:
+                print('Password verification failed! Try again!')
+    
+        else:  # This else is associated with the for-loop, not the if-statement
+            print(f"Password verification failed for {trials} times. Aborting!!")
+            sys.exit()
```
mitya57: Please exit with non-zero status. So, if some program calls `keyring` as a sub-process, it can realize that it failed.

Also, maybe the error message should go to stderr, not stdout.

keyring/cli.py
```diff
@@ -86,10 +86,26 @@ def do_get(self):
...
+
+# Rest of the code...
```
mitya57: What does this comment mean? Nothing will be executed after `sys.exit()`.

keyring/cli.py
```diff
@@ -86,10 +86,26 @@ def do_get(self):
...
-        password = self.input_password(
-            f"Password for '{self.username}' in '{self.service}': "
-        )
-        set_password(self.service, self.username, password)
+        trials = 3
+        for i in range(trials):
+            password = self.input_password(
+                f"Password for '{self.username}' in '{self.service}': ")
+    
+            reEnteredPassword = self.input_password(
```
mitya57: Please use PEP 8 compliant variable names, i.e. `reentered_password`.

mitya57 (changes requested):

keyring/cli.py
abastola0: Changed the code to exit with status 1 in case of error. Sent the error message to stdout.

keyring/cli.py
abastola0: Removed this.

keyring/cli.py
abastola0: changed the variable names.

mitya57 (changes requested):
Please also apply `black` formatting changes to make CI happy.

mitya57 (approved):

mitya57:
Thanks, approved. The final word is @jaraco's, he may have his own comments.

abastola0:
@mitya57 diffcov failed in this case. What do you think i might need to change.

mitya57:
Well, if you manage to write a test for this code, that would be awesome. If no, then I think we will have to ignore that warning.

keyring/cli.py
```diff
@@ -86,10 +86,28 @@ def do_get(self):
...
-            f"Password for '{self.username}' in '{self.service}': "
-        )
-        set_password(self.service, self.username, password)
+        trials = 3
+        for i in range(trials):
+            password = self.input_password(
+                f"password for '{self.username}' in '{self.service}': "
+            )
+
+            reentered_password = self.input_password(
```
jaraco, line 95: I've started work on tests and run into an issue. The `input_password` method is coded to accept the input from a pipe (if present). This change of expecting two passwords when it's piped in is unacceptable. Probably instead we'll want to hook this logic in at `getpass.getpass` instead of wrapping `input_password`.

jaraco: In #663, I'm working on adding some tests to capture at least the current expectation and hopefully serve as a foundation for testing this more sophisticated behavior. Interestingly, those tests didn't seem to capture the concern I mentioned above, but I expected them to, so I'll need to think about it a bit more before proceeding.

I've also worked on some refactorings in the `feature/password-verification` branch. I'd pushed them to this PR, but decided to leave them in their own branch for now while working out this issue.

abastola0: Sure. Please let me know if you need help with any other issues as well. I'll be more than happy to contribute.

mitya57: Maybe we should ask to re-enter password only if `sys.stdin.isatty()`?

abastola0: Yeah I think that works. I think it would be a good idea to skip verification for if not typed in explicitly. I can make this change if it's fine with @jaraco. I would imagine the piped password would already be visual to the user in some form so the verification would not make much sense in that case.

abastola0: Hi @jaraco. Any update on this?

jaraco: Sorry for the delay. Yes, please proceed with drafting an implementation that avoids the re-prompt for non-interactive cases.

jaraco:
This PR appears to be languishing, so I'm closing for now, but feel free to revive the conversation and we can re-open.

PR #663, opened 2023-11-19
Add test for do_set

jaraco:


jaraco:
It's annoying. The code seems to run fine on Python 3.11 without deprecation warnings but has problems on 3.12 with APIs having gone missing. What gives?

jaraco:
Aha: https://github.com/python/cpython/issues/100690

PR #665, opened 2023-12-12
feature: Allow listing of acct

VivienGiraud:
This PR add usage of listing accounts given an app name.

```python3
>>> keyring list MyApp
['GitHub', 'Microsoft']
```

Then you will be able to access an specific account using:
```python3
>>> keyring get MyApp Microsoft
MY_PASSWORD
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -87,6 +87,8 @@ or with ``--help`` as so::
     Password for 'username' in 'system':
     $ keyring get system username
     password
+    $ keyring list app_name
+    ['acct1', 'acct2', 'acct3']
 
 The command-line functionality is also exposed as an executable
 package, suitable for invoking from Python like so::
@@ -96,6 +98,8 @@ package, suitable for invoking from Python like so::
     Password for 'username' in 'system':
     $ python -m keyring get system username
     password
+    $ python -m keyring list app_name
+    ['acct1', 'acct2', 'acct3']
 
 Tab Completion
 --------------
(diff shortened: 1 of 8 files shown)


VivienGiraud:
Hi @mitya57 ,
Can you tell me if you could review this feature?
I working on an open source project that will rely on this functionality.
Have a pleasant day.

mitya57:
1. I would name the function `list_passwords` (in plural), not `list_password`.
2. Ideally we need support in all backends, not just in the macOS one.
3. The CI is failing. In particular, you forgot to import your function in `keyring/cli.py`.

Note: I'm not the chief maintainer of Keyring, @jaraco is, so he has the final word.

keyring/backend.py
```diff
@@ -100,6 +100,11 @@ def get_password(self, service: str, username: str) -> typing.Optional[str]:
         """Get password of the username for the service"""
         return None
 
+    @abc.abstractmethod
+    def list_password(self, service: str) -> list[str]:
```
jaraco, line 104: How about name it `list_accounts` or `list_users`, since "username" and "account" are the terms used for these fields and "passwords" aren't involved in the result.

README.rst
```diff
@@ -87,6 +87,8 @@ or with ``--help`` as so::
     Password for 'username' in 'system':
     $ keyring get system username
     password
+    $ keyring list app_name
```
jaraco, line 90: Probably this should mention "system" and not "app_name", because "system" is what's mentioned above.

jaraco, line 90: We should think carefully about the "command" here. It's called "list", but "list" could mean "list users" or "list systems" or "list users and systems". I wonder if we shouldn't instead allow a generic "list" that takes a string and matches any system:password where that string is found. Let's take some time to think about other use-cases other than exact matches of a system.

jaraco (changes requested):
Overall, this looks good. I have a couple of comments about naming and the design.

I'm also curious - what should this project think about the unimplemented backends? Ideally, a feature like this would be launched with support in Mac, Windows, and Linux. Would you be willing to explore an implementation that works in those backends?

Finally, this change should demand some tests. Let's make sure the test suite exercises this new expectation and functionality.

jaraco:
This PR appears to be languishing, so I'm going to close it for now. Feel free to revive the conversation and respond to the questions and we can re-open.

PR #666, opened 2024-02-10
simplify and fix typing of priority()

dimbleby:
as currently typed, mypy complains:
```
keyring/backends/fail.py:18: error: Signature of "priority" incompatible with supertype "KeyringBackend"  [override]
keyring/backends/fail.py:18: note:      Superclass:
keyring/backends/fail.py:18: note:          classproperty[float]
keyring/backends/fail.py:18: note:      Subclass:
keyring/backends/fail.py:18: note:          classproperty[int]
```
etc

but per https://peps.python.org/pep-0484/#the-numeric-tower, if you want `float | int`, `float`does just fine.  And now mypy is not unhappy with the typing of `priority()`.

### DIFF
--- a/keyring/_compat.py
+++ b/keyring/_compat.py
@@ -4,4 +4,4 @@ __all__ = ['properties']
 try:
     from jaraco.classes import properties  # pragma: no-cover
 except ImportError:
-    from . import _properties_compat as properties  # pragma: no-cover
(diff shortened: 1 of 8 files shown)


jaraco:
Aha. I addressed this issue in #667 and other typing issues in earlier commits, but I took a different approach. Let me resolve the conflicts and see what the new diff is.

PR #672, opened 2024-04-02
re-use ExceptionTrap from jaraco.context

jaraco:


jaraco:
I'm ignoring the diffcov failure because it is only for the errors file on the deprecated handler.

PR #678, opened 2024-04-16
feature: Add `get` options to cli for interface with `get_credential` and support JSON output

BakerNet:
Internally, `keyring` supports `get_credential`, which is useful for backends which can supply their own username.  This adds an exposed `get_credential` in the cli - `keyring --mode creds get [service] [OPTIONAL - username]`

This also adds support for JSON output type via `--output json` option

Help text for reference:

```text
  --output {plain,json}
                        Output format for 'get' operation. Default is 'plain'
  --mode {password,creds}
                        Mode for 'get' operation. 'password' requires a username and will return only the password.
                        'creds' does not require a username and will return both the username and password separated by
                        a newline. Default is 'password'
```

Example output:

```
 ➜  keyring --mode creds get https://example.com/
hello
world
 ➜  keyring --mode creds --output json get https://example.com/
{"username": "hello", "password": "world"}
 ➜  keyring --output json get https://example.com/ hello
{"password": "world"}
```

### DIFF
--- a/keyring/cli.py
+++ b/keyring/cli.py
@@ -82,7 +82,10 @@ class CommandLineTool:
 
     def _check_args(self):
         if self.operation:
-            if self.service is None or (self.operation != "getcreds" and self.username is None):
+            if self.operation == 'getcreds':
+                if self.service is None :
+                    self.parser.error(f"{self.operation} requires service")
+            elif self.service is None or self.username is None:
                 self.parser.error(f"{self.operation} requires service and username")
 
     def do_get(self):
--- a/keyring/cli.py
+++ b/keyring/cli.py
@@ -42,7 +42,21 @@ class CommandLineTool:
         self.parser.add_argument(
             "--disable", action="store_true", help="Disable keyring and exit"
         )
-        self.parser._operations = ["get", "set", "del", "getcreds", "diagnose"]
+        self.parser._get_modes = ["password", "creds"]
+        self.parser.add_argument(
+            "--get-mode",
+            choices=self.parser._get_modes,
+            dest="get_mode",
+            default="password",
+            help="""
+            Mode for 'get' operation.
+            'password' requires a username and will return only the password.
+            'creds' does not require a username and will return both the username and password separated by a newline.
+
+            Default is 'password'
+            """,
+        )
+        self.parser._operations = ["get", "set", "del", "diagnose"]
         self.parser.add_argument(
             'operation',
             choices=self.parser._operations,
@@ -82,24 +96,24 @@ class CommandLineTool:
 
     def _check_args(self):
         if self.operation:
-            if self.operation == 'getcreds':
+            if self.operation == 'get' and self.get_mode == 'creds':
                 if self.service is None:
                     self.parser.error(f"{self.operation} requires service")
             elif self.service is None or self.username is None:
                 self.parser.error(f"{self.operation} requires service and username")
 
     def do_get(self):
-        password = get_password(self.service, self.username)
-        if password is None:
-            raise SystemExit(1)
-        print(password)
-
-    def do_getcreds(self):
-        creds = get_credential(self.service, self.username)
-        if creds is None:
-            raise SystemExit(1)
-        print(f"username: {creds.username}")
-        print(f"password: {creds.password}")
+        if self.get_mode == 'creds':
+            creds = get_credential(self.service, self.username)
+            if creds is None:
+                raise SystemExit(1)
+            print(f"{creds.username}")
+            print(f"{creds.password}")
+        else:
+            password = get_password(self.service, self.username)
+            if password is None:
+                raise SystemExit(1)
+            print(password)
 
     def do_set(self):
         password = self.input_password(
--- a/keyring/cli.py
+++ b/keyring/cli.py
@@ -64,7 +64,7 @@ class CommandLineTool:
             dest="output_format",
             default="plain",
             help="""
-            Outpup format for 'get' operation.
+            Output format for 'get' operation.
 
             Default is 'plain'
             """,
@@ -127,7 +127,7 @@ class CommandLineTool:
             password = get_password(self.service, self.username)
             if password is None:
                 raise SystemExit(1)
-            crediential_dict['password'] = password
+            credential_dict['password'] = password
         if self.output_format == 'json':
             print(json.dumps(credential_dict))
         else:
--- a/keyring/cli.py
+++ b/keyring/cli.py
@@ -116,24 +116,30 @@ class CommandLineTool:
                 self.parser.error(f"{self.operation} requires service and username")
 
     def do_get(self):
-        credential_dict = {}
-        if self.get_mode == 'creds':
-            creds = get_credential(self.service, self.username)
-            if creds is None:
-                raise SystemExit(1)
-            credential_dict['username'] = creds.username
-            credential_dict['password'] = creds.password
-        else:
-            password = get_password(self.service, self.username)
-            if password is None:
-                raise SystemExit(1)
-            credential_dict['password'] = password
+        credential_dict = getattr(self, f'_get_{self.get_mode}')()
         if self.output_format == 'json':
             print(json.dumps(credential_dict))
         else:
             if 'username' in credential_dict.keys():
                 print(credential_dict['username'])
             print(credential_dict['password'])
+
+    def _get_cred(self):
+        creds = get_credential(self.service, self.username)
+        if creds is None:
+            raise SystemExit(1)
+        credential_dict = {}
+        credential_dict['username'] = creds.username
+        credential_dict['password'] = creds.password
+        return credential_dict
+
+    def _get_password(self):
+        password = get_password(self.service, self.username)
+        if password is None:
+            raise SystemExit(1)
+        credential_dict = {}
+        credential_dict['password'] = password
+        return credential_dict
 
     def do_set(self):
         password = self.input_password(
--- a/keyring/cli.py
+++ b/keyring/cli.py
@@ -42,7 +42,7 @@ class CommandLineTool:
         self.parser.add_argument(
             "--disable", action="store_true", help="Disable keyring and exit"
         )
-        self.parser._operations = ["get", "getcreds", "set", "del", "diagnose"] 
+        self.parser._operations = ["get", "set", "del", "getcreds", "diagnose"]
         self.parser.add_argument(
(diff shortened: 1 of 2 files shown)


keyring/cli.py
```diff
@@ -90,6 +94,13 @@ def do_get(self):
             raise SystemExit(1)
         print(password)
 
+    def do_getcreds(self):
+        creds = get_credential(self.service, self.username)
+        if creds is None:
+            raise SystemExit(1)
+        print(f"username: {creds.username}")
+        print(f"password: {creds.password}")
```
zanieb: Should we just print them newline separated without the prefix?

BakerNet: I am in favor of without prefix - I only included prefix because the only other command which outputs more than one item (`diagnose`) uses prefixes.

jaraco: What if the username or password contains a newline? Probably unlikely for both, but it's also maybe allowed?

FWIW, the `diagnose` interface is meant for human consumption so can be sloppy about syntax. I'm kinda leaning the same way for the rest of the keyring CLI API - meant primarily for human consumption. Still, no reason it can't provide a reasonably stable interface that a wrapper could rely upon.

I'm okay without the prefix as long as the help guides the user as to what they're getting from the output.

keyring/cli.py
```diff
@@ -41,7 +42,7 @@ def __init__(self):
         self.parser.add_argument(
             "--disable", action="store_true", help="Disable keyring and exit"
         )
-        self.parser._operations = ["get", "set", "del", "diagnose"]
+        self.parser._operations = ["get", "set", "del", "getcreds", "diagnose"]
```
jaraco: When naming things, I tend to separate words with something (space, dash, underscore). Here, I think dash may be appropriate.
```suggestion
        self.parser._operations = ["get", "set", "del", "get-creds", "diagnose"]
```

But then I notice we have `get` and `get-creds` and the asymetry is bothering me, and it makes me wonder if it should be `keyring get password` + `keyring get credential` (or `get pass` and `get creds`) or maybe `get` with `--creds` or `--password` (default). All of those options have more onerous transitional concerns, so I'm not loving any option.

I don't feel strongly about it, so happy to move forward with something pragmatic.

keyring/cli.py
zanieb: Is it allowed for the username or password to include a newline? I guess the keyring is arbitrary enough that it certainly could, but then it could also include `username: ` or `password: `.

The more complex option, I guess, is to output them base64 encoded.

keyring/cli.py
zanieb: `get-creds` seems like simplest transition for the API, I feel your pain though. Perhaps it'd be worth opening a tracking issue to consider a different API in some future major version. I think `get` with some sort of `--mode` might make the most sense. You could also just assume a single value is the service name and that the username is null and return the full credentials in that case? eek.

keyring/cli.py
jaraco: I looked at the tests and they do [test for difficult characters](https://github.com/jaraco/keyring/blob/35a3a384d333fc164949a6f85226209a922867bd/keyring/testing/backend.py#L35) on every keyring, so in theory they are supported. Still, I just wanted to flag it as a possibility.



> The more complex option, I guess, is to output them base64 encoded.

Oof. That wouldn't be user friendly. I'm thinking JSON would be better for safety, but that's not very friendly either. Maybe there should be an option to emit in an unambiguous form like JSON. (`--output json` (default 'plain'))?

zanieb: An output format enum seems reasonable.

Are the main users of the CLI humans? I presumed this was mostly used for programmatic interaction.

keyring/cli.py
BakerNet: > I think get with some sort of --mode might make the most sense.

This would be my personal vote over separate command.  Updated PR to add `--get-mode` with default `password`

Help text for reference:
```
  --get-mode {password,creds}
                        Mode for 'get' operation. 'password' requires a username and will return only the password.
                        'creds' does not require a username and will return both the username and password separated by
                        a newline. Default is 'password'
```

keyring/cli.py
```diff
@@ -56,6 +57,18 @@ def __init__(self):
...
             """,
         )
+        self.parser._output_formats = ["plain", "json"]
+        self.parser.add_argument(
+            "--output",
+            choices=self.parser._output_formats,
+            dest="output_format",
+            default="plain",
+            help="""
+            Outpup format for 'get' operation.
```
zanieb:
```suggestion
            Output format for 'get' operation.
```

keyring/cli.py
BakerNet: Added `--output` argument

Help text for reference:
```
  --output {plain,json}
                        Output format for 'get' operation. Default is 'plain'
```

Example output:
```
 ➜  keyring --get-mode creds --output json get https://example.com
{"username": "hello", "password": "world"}
 ➜  keyring --get-mode creds get https://example.com
hello
world
```

keyring/cli.py
zanieb: I don't really understand how we're going to use this downstream though. We'll need to try try `--get-mode creds` first and if it fails fall back to excluding the flag? Presumably there will be a different exit code? I guess we'll cache this so we don't need to do it on every request? but we have a lot of concurrent requests that make this difficult to do efficiently.

BakerNet: > I don't really understand how we're going to use this downstream though.

If no username on the index url, use `--get-mode creds` else use default.  After that, the cacheing by netloc should just work

zanieb: Yeah but we won't know if the user's `keyring` version supports this without trial and error.

BakerNet: > Yeah but we won't know if the user's keyring version supports this without trial and error.

You can tell this is the problem on the first call if the status code is `2` instead of `1` and don't try again after first `2` status.

```
➜ keyring get --get-mode creds https://example.com/ oauth2accesstoken
usage: keyring [-h] [-p KEYRING_PATH] [-b KEYRING_BACKEND] [--list-backends]
               [--disable] [--print-completion {bash,zsh,tcsh}]
               [{get,set,del,diagnose}] [service] [username]
keyring: error: unrecognized arguments: --get-mode creds https://example.com/
X➜   echo $?                       +
2
➜  keyring get https://example.com/ oauth2accesstoken
X➜  echo $?                       +
1
```

The user manually requests `keyring` use, so `uv` would already be failing for them if no username provided.  This is just a different failure mode.

`if status code 2, warn user to either use username in index-url or update to keyring version X`

BakerNet: Err... I think I see an issue - for index-urls which don't need keyring (PyPi) - you'd hit the status code `2` when keyring requested but don't know whether you need to warn the user about it.

BakerNet: > I guess we'll cache this so we don't need to do it on every request? but we have a lot of concurrent requests that make this difficult to do efficiently.

On this point, you're using a mutex on the keyring cache, so the first time a host fails the check, you're returning None for all other checks on that host (keyring only running once per host).  The only way you'll run into new efficiency issues is if there are a large number of unique index-url hosts involved.

This seems unlikely - but if you want to avoid it, you could check the exit code of the command and skip subsequent `keyring` calls if status code was `2`.

zanieb: 👍  Sorry I didn't mean to distract, unless usage in `uv` helps inform the design here I don't think we need to discuss it too much.

BakerNet: @zanieb We're kind of off-topic to discussion of this PR - can move to draft example PR on `uv` here:
https://github.com/astral-sh/uv/pull/3081

keyring/cli.py
jaraco: > Are the main users of the CLI humans? I presumed this was mostly used for programmatic interaction.

I'm unsure. In my personal usage, I always use the CLI as a human and always use the library for programming, but that's because I'm always coding in Python. It's my impression that my experience has been the most common experience for a long time but that may be changing, especially with the introduction of keyring to homebrew and the use of it by tools like uv.

jaraco: > Added `--output` argument
> 
> Help text for reference:
> 
> ```
>   --output {plain,json}
>                         Output format for 'get' operation. Default is 'plain'
> ```
> 
> Example output:
> 
> ```
>  ➜  keyring --get-mode creds --output json get https://example.com
> {"username": "hello", "password": "world"}
>  ➜  keyring --get-mode creds get https://example.com
> hello
> world
> ```

I'm liking this a lot! Thanks. Is it possible to use `--mode` (since `get` is implied by the command)?

BakerNet: Done

keyring/cli.py
```diff
@@ -81,14 +109,31 @@ def run(self, argv):
 
     def _check_args(self):
         if self.operation:
-            if self.service is None or self.username is None:
+            if self.operation == 'get' and self.get_mode == 'creds':
+                if self.service is None:
+                    self.parser.error(f"{self.operation} requires service")
+            elif self.service is None or self.username is None:
                 self.parser.error(f"{self.operation} requires service and username")
```
jaraco: This 6 line function now has a cyclomatic complexity of 5, up from 3 in the prior implementation. I'm going to see if it's possible to keep it lower.

jaraco: Done in 33bae95.

jaraco:
Releasing as v25.2.0.

keyring/cli.py
jfly: (sorry I'm late to the party)

> > Are the main users of the CLI humans? I presumed this was mostly used for programmatic interaction.
>
> I'm unsure. In my personal usage, I always use the CLI as a human and always use the library for programming, but that's because I'm always coding in Python. It's my impression that my experience has been the most common experience for a long time but that may be changing, especially with the introduction of keyring to homebrew and the use of it by tools like uv.

If another datapoint is useful: pip has support for invoking `keyring` both [via python import](https://pip.pypa.io/en/stable/topics/authentication/#using-keyring-s-python-module) and [as a subprocess](https://pip.pypa.io/en/stable/topics/authentication/#using-keyring-as-a-command-line-application), and the author ["recommend[s] you use the `subprocess` provider](https://github.com/pypa/pip/issues/12423#issuecomment-1872936029).

edit: BTW, thanks @BakerNet for this feature! It allows us to bring pip's keyring subprocess provider up to feature parity with pip's keyring import provider: https://github.com/pypa/pip/pull/12748

Issue #668, opened 2024-03-07
Support for empty string usernames is inconsistent


djkawa:
Describe the bug:

Storing a password with an empty string "" username in windows 64, returns None with python 3.12.

To Reproduce:

All on windows 64, set up an environment with python 3.12 and keyring 24.3.1. Set a password with an empty string username "". Retrieve the password -> None.

Expected behavior
Doing the same thing with python 3.11, the password is retrieved. 

Environment

    Windows 64
    Python 3.12.2
    Keyring 24.3.1

djkawa:
Describe the bug:

Storing a password with an empty string "" username in windows 64, returns None with python 3.12.

To Reproduce:

All on windows 64, set up an environment with python 3.12 and keyring 24.3.1. Set a password with an empty string username "". Retrieve the password -> None.

Expected behavior
Doing the same thing with python 3.11, the password is retrieved. 

Environment

    Windows 64
    Python 3.12.2
    Keyring 24.3.1

jaraco:
On my Windows system, I see consistent behavior across Python versions:

```
 ~ # py -3.11 -m pip-run keyring -- -m keyring set system ''
Password for '' in 'system': 
 ~ # py -3.11 -m pip-run keyring -- -m keyring get system ''
 ~ # py -3.12 -m pip-run keyring -- -m keyring get system ''
```

I can confirm in credential manager that the password is being stored with an empty username. I'm unsure why it's not able to be retrieved.

jaraco:
I'm not confident that empty usernames was ever supported. I searched the code and there are no tests guaranteeing that empty usernames work, so if they happened to work, it was a lucky accident. In bc34083, I added such a test, and it's failing right away (on Windows only), seemingly due to a failure to delete the password on cleanup.

More troubleshooting is called for. Can you tell me more about your use-case? How long have you been using empty usernames? Do you know why I've been unable to replicate your findings? What version of Python 3.11 do you have? Can you confirm that you have the latest keyring on both Python 3.11 and 3.12?

djkawa:
Hi Jason,Thanks for your answers.

I use conda forge to create my environments under windows 10 64 bits.

Python 3.11 is 3.11.8
Python 3.12 is 3.12.2

[...]

PR #687, opened 2024-08-01
Deprecate empty usernames

jaraco:
- **Add a test capturing the missed expectation. Ref #668.**
- **Mark the test as xfail where it's failing (Windows).**
- **Deprecate setting a password for an empty username.**
- **Extract method for registration.**
- **When constructing KeyringBackend subclasses, wrap set_password to enforce non-empty usernames.**


jaraco:
Tests on diffcov are failing, but only because the coverage reports aren't aggregated (jaraco/skeleton#130).

mitya57:
> or comment in the issue

There will be an issue link generated, right?

jaraco:
> > or comment in the issue
> 
> There will be an issue link generated, right?

Yes. It will link to #668 in the changelog (from the 668.feature.rst file).

mitya57 (approved):

PR #689, opened 2024-08-10
add a couple of type annotations

dimbleby:
I ran into `untyped-call` warnings using the `SimpleCredential` - so here is a pull request adding annotations in `credentials.py`.

(There is a lot of unannotated code in this repository, I have no plans to do anything about the rest of it)

### DIFF
--- a/keyring/credentials.py
+++ b/keyring/credentials.py
@@ -8,8 +8,7 @@ class Credential(metaclass=abc.ABCMeta):
     """Abstract class to manage credentials"""
 
     @abc.abstractproperty
-    def username(self) -> str | None:
-        return None
+    def username(self) -> str: ...
 
     @abc.abstractproperty
     def password(self) -> str: ...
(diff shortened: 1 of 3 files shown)


keyring/credentials.py
```diff
@@ -6,27 +8,27 @@ class Credential(metaclass=abc.ABCMeta):
...
-    def password(self):
+    def password(self) -> str | None:
         return None
 
 
 class SimpleCredential(Credential):
     """Simple credentials implementation"""
 
-    def __init__(self, username, password):
+    def __init__(self, username: str | None, password: str | None):
```
jaraco: Do we know that there are cases expecting that username and password can be null? There's been some work recently to enforce that the username cannot be null when setting a password. The `set_password` requires `str` for both password and username. I'm thinking the credential should not allow null values here.

dimbleby: Eg https://github.com/jaraco/keyring/blob/ccdb048f9ceb7d6451b7bd34106fec84449e2a07/keyring/cli.py#L141

jaraco: Okay. So maybe username should allow for null. What about password?

dimbleby: I don't mind either way, happy for you to say what the intended usage is. 

I see that you use a null username - and also that [your ABC defaults to `None` for both properties](https://github.com/jaraco/keyring/blob/ccdb048f9ceb7d6451b7bd34106fec84449e2a07/keyring/credentials.py#L5-L14) - and so I err on the side of caution

jaraco: > Eg
> 
> https://github.com/jaraco/keyring/blob/ccdb048f9ceb7d6451b7bd34106fec84449e2a07/keyring/cli.py#L141

I'm actually leaning toward creating an `AnonymousCredential` to represent a credential without a username. That would help prevent nullable properties, so `SimpleCredential` could insist on set properties. Let's start with `SimpleCredential` expecting a password.

I can take it from here. Thanks for the contrib and the clarifications.

Issue #692, opened 2024-09-17
Posible regresion in 25.4.0  breaking  keyrings-google-artifactregistry-auth


07pepa:
**Describe the bug** 
Posible regresion in  25.4.0 causing ValueError: Anonymous credential has no username with keyrings-google-artifactregistry-auth 

rolling back 25.3.0 works 


**To Reproduce**

1. install uv
2. uv tool install "keyring" --with keyrings-google-artifactregistry-auth
3. have dependecies managed by uv in toml
4. have python artifact registry on GCP
5. have toml set up correctly like
```toml
[tool.uv]
package = false
index-url = "https://oauth2accesstoken@europe-python.pkg.dev/<REDACTED>/simple/"
keyring-provider = "subprocess"
```

**Expected behavior**
authenticates and works

**Observed  behavior**
```
.811   File "/root/.local/bin/keyring", line 8, in <module>                                                                                                                                                                       
1.812     sys.exit(main())                                                                                                                                                                                                         
1.812              ^^^^^^                                                                                                                                                                                                          
1.813   File "/root/.local/share/uv/tools/keyring/lib/python3.12/site-packages/keyring/cli.py", line 219, in main
1.813     return cli.run(argv)
1.813            ^^^^^^^^^^^^^
1.813   File "/root/.local/share/uv/tools/keyring/lib/python3.12/site-packages/keyring/cli.py", line 120, in run
1.813     return method()
1.813            ^^^^^^^^
1.813   File "/root/.local/share/uv/tools/keyring/lib/python3.12/site-packages/keyring/cli.py", line 132, in do_get
1.814     getattr(self, f'_emit_{self.output_format}')(credential)
1.814   File "/root/.local/share/uv/tools/keyring/lib/python3.12/site-packages/keyring/cli.py", line 140, in _emit_plain
1.814     if credential.username:
1.814        ^^^^^^^^^^^^^^^^^^^
1.814   File "/root/.local/share/uv/tools/keyring/lib/python3.12/site-packages/keyring/credentials.py", line 39, in username
1.814     raise ValueError("Anonymous credential has no username")
1.814 ValueError: Anonymous credential has no username
```


**Environment**

- OS: latest docker  python:3.12-slim

```
$ uv tool list | grep keyring
keyring v25.4.0

$ uv run keyring --list-backends
keyring.backends.chainer.ChainerBackend (priority: -1)
keyrings.gauth.GooglePythonAuth (priority: 9)
[...]

PR #693, opened 2024-09-17
Fix plain emit with Anonymous Credentials #692

daaniam:


Tommaso-FLYR:
thank you! This broke our CI :/

daaniam:
@Tommaso-FLYR, do you have an estimated timeframe for when it might be fixed? We use it without a pinned version in a few Dockerfiles as a UV tool. Thank you.

Tommaso-FLYR:
> @Tommaso-FLYR, do you have an estimated timeframe for when it might be fixed? We use it without a pinned version in a few Dockerfiles as a UV tool. Thank you.

well I'm not a maintainer here so I can't answer to that. However we probably have the same setup (dockerfile + uv)

This is how I fixed it:
```diff
- RUN uv pip install keyring keyrings.google-artifactregistry-auth && \
+ RUN uv pip install keyring==25.3.0 keyrings.google-artifactregistry-auth && \
    uv pip install --keyring-provider subprocess -r requirements.txt && \
    uv pip uninstall keyring keyrings.google-artifactregistry-auth
```

jaraco:
Per #694, the issue also affects `_emit_json`. Since #695 addresses both concerns, I'm going to work from that one. Regardless, thanks for the contrib, as it re-enforces the approach proposed there.

Issue #698, opened 2024-10-04
WinVaultKeyring.get_credential with non-existent username returns credential of other user


JamieBeverley:
**Describe the bug**

`WinVaultKeyring.get_credential("service", "<username_dne>")` with non-existent username can return credentials of another username. 

**To Reproduce**
```python
import keyring
from keyring.backends.Windows import WinVaultKeyring

keyring.set_keyring(WinVaultKeyring())

keyring.set_password(
    service_name='service1',
    username='user1',
    password='password1'
)

keyring.set_password(
    service_name='service1',
    username='user2',
    password='password2'
)

# As expected:
print(keyring.get_credential(service_name="service1",username="user1").password)
print(keyring.get_credential(service_name="service1",username="user2").password)

# I think as expected (defaults most recent? not 100% confident of expected)
# behavior here
print(keyring.get_credential(service_name="service1",username=None).password)

# I think we should expect `None`?
print(keyring.get_credential(service_name="service1",username="nobody!").password)
```
Output:
```
password1
password2
password2
password2
```

**Expected behavior**

`keyring.get_credential("service1", "<username dne>")` should return `None` when credential for `service1` - `<username dne>` doesn't exist.

**Environment**

- OS: Windows
- keyring: 2.4.1
- backend: `WinVaultKeyring`

**Additional context**

- `get_password` does not have the same issue (there is a check that the `username` parameter matches the credential pulled from wincred which should perhaps be added to `get_credential` as well.
- Will share a PR for consideration shortly to illustrate more precisely

Possibly tangential to some other windows-labelled issues - eg #664

jaraco:
Fixed in #699.

PR #699, opened 2024-10-04
Bugfix: WinVaultKeyring.get_credential with non-existent username returns credential of other user (#698)

JamieBeverley:
Regarding #698

### DIFF
--- a/keyring/backends/Windows.py
+++ b/keyring/backends/Windows.py
@@ -164,7 +164,8 @@ class WinVaultKeyring(KeyringBackend):
             res = self._get_password(self._compound_name(username, service))
         # get any first password under the service name
         if not res:
-            res = self._get_password(service)
-            if not res:
-                return None
+            cred = self._get_password(service)
+            res = cred if username is None or username == cred["UserName"] else None
+        if not res:
+            return None
         return SimpleCredential(res['UserName'], res.value)
--- a/keyring/backends/Windows.py
+++ b/keyring/backends/Windows.py
@@ -94,16 +94,20 @@ class WinVaultKeyring(KeyringBackend):
         return f'{username}@{service}'
 
     def get_password(self, service, username):
-        # first attempt to get the password under the service name
-        res = self._get_password(service)
-        if not res or res['UserName'] != username:
-            # It wasn't found so attempt to get it with the compound name
-            res = self._get_password(self._compound_name(username, service))
-        if not res:
-            return None
-        return res.value
+        res = self._resolve_credential(service, username)
+        return res and res.value
 
-    def _get_password(self, target):
+    def _resolve_credential(
+        self, service: str, username: str | None
+    ) -> DecodingCredential | None:
+        # first attempt to get the password under the service name
+        res = self._read_credential(service)
+        if not res or username and res['UserName'] != username:
+            # It wasn't found so attempt to get it with the compound name
+            res = self._read_credential(self._compound_name(username, service))
+        return res
+
+    def _read_credential(self, target):
         try:
             res = win32cred.CredRead(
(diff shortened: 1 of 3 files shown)


keyring/testing/backend.py
```diff
@@ -181,3 +181,20 @@ def test_new_with_properties(self):
...
+        assert cred is None
+
+        password_1 = 'password1'
+        password_2 = 'password2'
+        self.set_password(service, 'user1', password_1)
+        self.set_password(service, 'user2', password_2)
+
+        assert keyring.get_credential(service, "user1").password == password_1
+        assert keyring.get_credential(service, "user2").password == password_2
+
```
JamieBeverley: Question/Note:
Considered adding this additional assertion too (see [previous commit)](https://github.com/jaraco/keyring/pull/699/commits/27d0e0ea3ea75ad7400e9d1d42bcbbbe088aaa01) but it was failing intermittently on linux.

Sometimes `keyring.get_credential(service, None).password ` would equate to `password_2` and sometimes to `password_1`.

Sticking a `time.sleep(1)` between the 2 `set_password` calls above helped it reliably pass so perhaps there's some async/timing thing happening with dbus (can possibly look into that deeper but a bit out of my depths w/ dbus)

```suggestion
        assert keyring.get_credential(service, "user2").password == password_2
         # Most recent entered
        assert keyring.get_credential(service, None).password == password_2
```

jaraco: Probably that additional assertion should be a separate test. Better to test one thing (the thing that matches the function's name). I'll use your comment to file a separate issue to track this concern (#700).

keyring/backends/Windows.py
```diff
@@ -164,7 +164,8 @@ def get_credential(self, service, username):
             res = self._get_password(self._compound_name(username, service))
         # get any first password under the service name
```
jaraco: This comment no longer matches the implementation.

keyring/backends/Windows.py
```diff
@@ -164,7 +164,8 @@ def get_credential(self, service, username):
...
         if not res:
-            res = self._get_password(service)
-            if not res:
-                return None
+            cred = self._get_password(service)
+            res = cred if username is None or username == cred["UserName"] else None
+        if not res:
+            return None
```
jaraco: I noticed these two lines were moved to the top level, but they are only relevant in this branch. Probably it makes sense not to change this behavior.

```suggestion
            if not res:
                return None
```

keyring/backends/Windows.py
```diff
@@ -164,7 +164,8 @@ def get_credential(self, service, username):
             res = self._get_password(self._compound_name(username, service))
         # get any first password under the service name
         if not res:
-            res = self._get_password(service)
-            if not res:
-                return None
+            cred = self._get_password(service)
```
jaraco: nitpick: the variable `cred` was used to represent the result from an API call (`_get_password`/`win32cred.CredRead`). Now `cred` is being used for the same purpose. I see why you did it, though. You needed a different name to represent the intermediate result to allow for `res` to represent the value when exiting the branch. Not sure there's anything to be done here.

keyring/backends/Windows.py
jaraco: Fixed in 2f3f4ec.

jaraco (approved):
This looks good. I'm going to do a bit of subsequent work to reorganize.

jaraco:
I've always bemoaned that `get_credential` and `get_password` both had logic for resolving a credential. With 5e4b99e, that behavior is now unified and the tests still pass.

JamieBeverley:
Thanks @jaraco - appreciate the comments + improvements!

PR #704, opened 2025-01-02
Remove unused dependency jeepney

WhyNotHugo:
jeepney>=0.4.2 is listed as a dependency, but the README indicates that the jeepney backed was moved into a separate project. The dependency is not used directly any more either.

Remove the unused dependency completely.

See: https://gitlab.com/takluyver/keyring_jeepney

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,3 +1,11 @@
+v25.7.0
+=======
+
+Bugfixes
+--------
+
+- Removed unused dependency ``jeepney``.
+
 v25.6.0
 =======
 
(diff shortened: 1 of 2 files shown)


mitya57:
Okay, we depend on `SecretStorage>=3.2` on the previous line, and SecretStorage depends on `jeepney>=0.6`, so this does not change much anyway.

mitya57 (approved):

WhyNotHugo:
You're quite right. I was under the impression that it used `dbus-python`.

mitya57:
SecretStorage used to use dbus-python, but in version 3.0.0 it was ported to Jeepney, which is much easier to install.

Issue #283, opened 2017-06-30
"No recommended backend" under Tox with all known dependencies installed


Alan-R:
When running my tests under Tox, I get this well-known message:

> RuntimeError: No recommended backend was available. Install the keyrings.alt package if you want to use the non-recommended backends. See README.rst for details.

When I run the same tests under nose (outside of tox) everything is happy...

Here's what's in my tox.ini:
```
[tox]
site-packages=True
[testenv]
site-packages=True
deps=
    dbus-python==1.2.4
    SecretStorage==2.3.1
    keyring==10.4.0
```

Here's what Tox says is installed:
```
python installed: asn1crypto==0.22.0,cffi==1.10.0,coverage==4.4.1,cryptography==1.9,dbus-python==1.2.4,enum34==1.1.6,idna==2.5,ipaddress==1.0.18,keyring==10.4.0,naas-base==1.0.3,nose==1.3.7,pycparser==2.17,SecretStorage==2.3.1,six==1.10.0
python finish: envreport after 0.30 seconds
python start: runtests
  removing /home/alanr/Charter/pypackage/naas/naas-base/.tox/python/tmp
python runtests: PYTHONHASHSEED='1153821658'
python runtests: commands[0] | nosetests --with-coverage --cover-erase -vw .
setting PATH=/home/alanr/Charter/pypackage/naas/naas-base/.tox/python/bin:/home/alanr/Charter/pypackage/naas/naas-base/venv/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/snap/bin:/sbin:/usr/sbin:/home/alanr/bin:/sbin:/usr/sbin
  /home/alanr/Charter/pypackage/naas/naas-base$ /home/alanr/Charter/pypackage/naas/naas-base/.tox/python/bin/nosetests --with-coverage --cover-erase -vw .
```

jaraco:
It looks like you're expecting to get the secretservice backend. You can troubleshoot by running `.tox/python/bin/python` and then inspecting that backend priority:

```
>>> import keyring
>>> keyring.backends.SecretService.Keyring.priority
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/keyring/util/properties.py", line 22, in __get__
    return self.fget.__get__(None, owner)()
[...]

PR #724, opened 2025-08-06
Add instructions for using Keyring with tox for pip credentials

dachucky:
Adds instructions to the Readme for how to deal with issues when using Keyring with tox for pip credentials. See #283 for context.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -382,16 +382,17 @@ The following is a complete transcript for installing keyring on a Ubuntu 18:04 
 Using Keyring with tox on Linux systems for pip credentials 
 =============================================================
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -379,6 +379,21 @@ The following is a complete transcript for installing keyring on a Ubuntu 18:04
   >>> keyring.get_password("system", "username")
   'password'
 
+Using Keyring with tox on Linux systems for pip credentials
```
jaraco: I don't think this guidance is specific to pip. It's not even specific to Linux, although Linux users are more likely to be caught by it. I'll generalize the guidance a bit.

jaraco:
Thanks for the contrib!

PR #740, opened 2026-02-13
Fix ChainerBackend.delete_password to try all backends

### LINKED ISSUE #697: ChainerBackend not deleting password from lower priority backend
**Describe the bug**
Using kdewallet I can see old keyring passwords listed in "Secret Service". New keyring passwords are listed under the "system" passed to set_password. I assume this is because I previously did not have `keyring.backends.kwallet.DBusKeyring` installed.
```
$ python3.11 -m keyring --list-backends
keyring.backends.chainer.ChainerBackend (priority: 10)
keyring.backends.kwallet.DBusKeyring (priority: 5.1)
keyring.backends.SecretService.Keyring (priority: 5)
```
`$ python3.11 -m keyring get system username` correctly finds the old passwords. `$ python3.11 -m keyring del system username` fails as follows:
```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/jim/.local/lib/python3.11/site-packages/keyring/__main__.py", line 4, in <module>
    cli.main()
  File "/home/jim/.local/lib/python3.11/site-packages/keyring/cli.py", line 216, in main
    return cli.run(argv)
           ^^^^^^^^^^^^^
  File "/home/jim/.local/lib/python3.11/site-packages/keyring/cli.py", line 120, in run
    return method()
           ^^^^^^^^
  File "/home/jim/.local/lib/python3.11/site-packages/keyring/cli.py", line 157, in do_del
    delete_password(self.service, self.username)
  File "/home/jim/.local/lib/python3.11/site-packages/keyring/core.py", line 73, in delete_password
    get_keyring().delete_password(service_name, username)
  File "/home/jim/.local/lib/python3.11/site-packages/keyring/backends/chainer.py", line 63, in delete_password
    return keyring.delete_password(service, username)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jim/.local/lib/python3.11/site-packages/keyring/backends/kwallet.py", line 150, in delete_password
    raise PasswordDeleteError("Password not found")
keyring.errors.PasswordDeleteError: Password not found
```

**Expected behavior**
I would expect to be able to delete any password that I can read.

**Environment**
- OS: openSUSE 15.6

```
$ python3.11 -m pip list | grep keyring
keyring                       25.4.1

$ python3.11 -m keyring --list-backends
keyring.backends.chainer.ChainerBackend (priority: 10)
keyring.backends.kwallet.DBusKeyring (priority: 5.1)
keyring.backends.SecretService.Keyring (priority: 5)
```

veeceey:
## Summary

Fixes #697 where `ChainerBackend.delete_password()` was only trying the first backend instead of iterating through all backends in the chain.

The issue occurred when a password existed in a lower priority backend but not in the highest priority one. The `delete_password` method would try only the first (highest priority) backend, get a `PasswordDeleteError`, and fail immediately.

## Changes

- Updated `ChainerBackend.delete_password()` to iterate through all backends, similar to how `get_password()` works
- Now catches `PasswordDeleteError` exceptions and continues trying other backends
- If all backends fail to delete the password, raises the last `PasswordDeleteError` encountered
- Added comprehensive test coverage for the new behavior

## Testing

Added two new test cases:
1. `test_delete_password_tries_all_backends` - Verifies that deletion succeeds when password exists only in a lower priority backend
2. `test_delete_password_raises_if_all_backends_fail` - Verifies that appropriate error is raised when password doesn't exist in any backend

All existing tests continue to pass.

### DIFF
--- a/keyring/backends/chainer.py
+++ b/keyring/backends/chainer.py
@@ -59,18 +59,14 @@ class ChainerBackend(backend.KeyringBackend):
                 pass
 
     def delete_password(self, service, username):
-        errors = []
         for keyring in self.backends:
             try:
                 return keyring.delete_password(service, username)
             except NotImplementedError:
                 pass
-            except PasswordDeleteError as e:
-                errors.append(e)
-        
-        # If we tried all backends and none succeeded, raise the last error
-        if errors:
-            raise errors[-1]
+            except PasswordDeleteError:
+                if keyring == self.backends[-1]:
+                    raise
 
     def get_credential(self, service, username):
         for keyring in self.backends:
--- a/keyring/backends/chainer.py
+++ b/keyring/backends/chainer.py
@@ -5,6 +5,7 @@ discover passwords in each.
 
 from .. import backend
 from ..compat import properties
+from ..errors import PasswordDeleteError
 from . import fail
 
 
@@ -58,11 +59,18 @@ class ChainerBackend(backend.KeyringBackend):
                 pass
 
     def delete_password(self, service, username):
+        errors = []
         for keyring in self.backends:
             try:
                 return keyring.delete_password(service, username)
             except NotImplementedError:
                 pass
+            except PasswordDeleteError as e:
+                errors.append(e)
+        
+        # If we tried all backends and none succeeded, raise the last error
+        if errors:
+            raise errors[-1]
 
     def get_credential(self, service, username):
         for keyring in self.backends:
--- a/tests/backends/test_chainer.py
+++ b/tests/backends/test_chainer.py
@@ -36,6 +36,7 @@ def delete_test_keyrings(monkeypatch):
     """
     Fixture that creates backends where passwords exist in lower priority backend.
     """
+
     class HighPriorityKeyring(backend.KeyringBackend):
         priority = 2
         storage = {}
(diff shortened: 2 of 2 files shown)


keyring/backends/chainer.py
```diff
@@ -58,11 +59,18 @@ def set_password(self, service, username, password):
                 pass
 
     def delete_password(self, service, username):
+        errors = []
```
mitya57: Why do we need a list when you use only the last item anyway?

You could just have a single variable and it will always have the last error.

Alternatively, you can avoid the variable at all and do something like:
```python
if keyring == self.backends[-1]:
    raise
```

mitya57 (changes requested):

keyring/backends/chainer.py
veeceey: Good point, the list was unnecessary. Switched to the `if keyring == self.backends[-1]: raise` approach - much cleaner. Updated in 720aef6.

mitya57 (approved):

veeceey:
Pushed a fix for the Windows CI mypy failure — the `storage: dict = {}` class-level type annotation was triggering mypy's `annotation-unchecked` note. Removed the type annotation since the `__init__` method sets `self.storage` anyway.

The other two CI failures (Windows 3.9 pip install failure and pypy3.10 timeout/cancellation) appear to be transient infrastructure issues unrelated to these changes.

PR #741, opened 2026-02-13
Fix case-sensitive username comparison in WinVaultKeyring

### LINKED ISSUE #736: WinVaultKeyring backend does not always accept case-insensitive usernames
**Describe the bug**
The WinVaultKeyring backend does not always accept case-insensitive usernames, even when Win32Cred does.

The following function causes this inconsistency:
https://github.com/jaraco/keyring/blob/38c040133559682902f25fe96496756ee6849820/keyring/backends/Windows.py#L102-L110

The code checks if the service exists in the Windows Credential Locker with an exact, case-sensitive match of the username (`res['UserName'] != username`). If it's not found, either because it doesn't exist or because the casing is different, it looks up the compound name `{username}@{service}`, which completely skips username comparison.

**To Reproduce**
```powershell
> keyring set test USER
Password for 'USER' in 'test': 
> keyring get test USER
PASS
> keyring get test user
> keyring set test user
Password for 'user' in 'test': 
> keyring get test user
pass
> keyring get test USER
PASS
> keyring get test usER
PASS
> keyring set test usER
Password for 'usER' in 'test': 
> keyring get test usER
paSS
> keyring get test USER
pass
> keyring get test user
pass
```

**Expected behavior**
WinVaultKeyring should use consistent case sensitivity.

**Environment**

- OS: Windows

```
> pip list | Select-String "keyring"
keyring           25.7.0
> keyring --list-backends
keyring.backends.chainer.ChainerBackend (priority: -1)
keyring.backends.Windows.WinVaultKeyring (priority: 5)
```

veeceey:
Fixes #736

The WinVaultKeyring backend was comparing usernames with strict case-sensitive equality (`res['UserName'] != username`), which doesn't match how Windows itself handles credentials — Win32Cred treats usernames case-insensitively.

This caused confusing behavior where storing a password for `USER` and then retrieving it as `user` would silently fail, and subsequent `set_password` calls with different casing would create orphaned compound entries.

**Changes:**

- Added a `_username_match` helper that does case-insensitive comparison, handling `None` values correctly
- Applied it in `_resolve_credential` so `get_password` / `get_credential` find credentials regardless of username casing
- Applied it in `set_password` so updating a password for the same user with different casing just overwrites in place instead of unnecessarily creating a compound `{username}@{service}` entry
- Applied it in `delete_password` so deletion works regardless of username casing
- Added unit tests covering all the above scenarios (mocked, so they run on all platforms)

**Test results:**

All existing tests continue to pass. The new tests in `TestUsernameMatch` and `TestWinVaultCaseInsensitive` cover the case-insensitive matching behavior across get, set, delete, and credential resolution.

### DIFF
--- a/keyring/backends/Windows.py
+++ b/keyring/backends/Windows.py
@@ -62,6 +62,13 @@ class DecodingCredential(dict):
             return decoded_cred_utf8
 
 
+def _username_match(a: str | None, b: str | None) -> bool:
+    """Compare usernames case-insensitively to match Windows behavior."""
+    if a is None or b is None:
+        return a is b
+    return a.lower() == b.lower()
+
+
 class WinVaultKeyring(KeyringBackend):
     """
     WinVaultKeyring stores encrypted passwords using the Windows Credential
@@ -104,7 +111,7 @@ class WinVaultKeyring(KeyringBackend):
     ) -> DecodingCredential | None:
         # first attempt to get the password under the service name
         res = self._read_credential(service)
-        if not res or username and res['UserName'] != username:
+        if not res or username and not _username_match(res['UserName'], username):
             # It wasn't found so attempt to get it with the compound name
             res = self._read_credential(self._compound_name(username, service))
         return res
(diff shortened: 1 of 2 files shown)


mitya57 (approved):

veeceey:
The Windows test failures look like they might be pre-existing/infra-related rather than caused by this change. Happy to investigate further if needed though.

PR #742, opened 2026-02-21
Address jaraco/keyring#709

### LINKED ISSUE #709: Worth reducing the number of dependencies?
First, thanks for building keyring!

Would it be possible to reduce the number of dependencies in any way?

We are in a weird situation where we would consider `keyring` to improve the security of a system by using the OS keychain, but at the same time, installing `keyring` means installing 9 additional dependencies, which increases the attack surface. (And potentially also increases the start-up time and could result in a conflict of indirect dependencies.)

So, a few questions:
- Why depend on `jaraco-classes` if the relevant code is already copied over in `keyring/copat/properties.py`?
- From `jaraco.functools` we only use `once`. Could we copy this over in the same way `properties.py` is copied? And the same for `ExceptionTrap` and `suppress` from `jaraco.context`?
- Both `jaraco-classes` and `jaraco-functools` depend on `more-itertools`. But in the end `more-itertools` is not used at all. (unless I miss something). Having a dependency installed that is not used at all doesn't sound great for a security related package.
- Less problematic, but given that I'd prefer not to load plug-ins, a dependency on `importlib_metadata` would also not be needed.

What do you think? Are these valid concerns that could be addressed to some point? Would you accept contributions? Or would it be better to fork the project in this case, or look for an alternative?

andraxin:
Closes #709, #704.

The new `backend` extra may be worth a mention in the README...

### DIFF
--- a/keyring/backend.py
+++ b/keyring/backend.py
@@ -12,12 +12,10 @@ import operator
 import os
 import typing
 import warnings
-
-from jaraco.context import ExceptionTrap
-from jaraco.functools import once
+from functools import lru_cache as once
 
 from . import credentials, errors, util
-from .compat import properties
+from .compat import ExceptionTrap, properties
 from .compat.py312 import metadata
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


andraxin:
@jaraco
I cannot seem to be able to edit **Reviewers** here.
How do I request a review?

PR #745, opened 2026-03-09
Remove obsolete call to win32_ver

ncoghlan:
platform.win32_ver is a deceptively complex function, and its output is no longer used here, so omit the call entirely.

(I encountered this in a test environment that had all unexpected subprocess invocations blocked via `pytest-subprocess`, and it picked up the attempted call to `ver` from `platform.win32_ver()`)

### DIFF
--- a/keyring/util/platform_.py
+++ b/keyring/util/platform_.py
@@ -4,7 +4,6 @@ import platform
 
 
 def _data_root_Windows():
-    release, version, csd, ptype = platform.win32_ver()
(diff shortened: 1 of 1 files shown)


mitya57 (approved):

ncoghlan:
The pypy3.10 test execution failure doesn't look related to the PR, it looks like something went wrong in the test execution environment. Might be worth just rerunning the failed checks?

Issue #644, opened 2023-07-31
two keyring password dialogs are triggered


bwagner:
**Describe the bug**

If name1 is a **substring** of name2, **two** dialogs are triggered to retrieve either of their passwords.
The first dialog reads:
`python3.11 wants to use your confidential information stored in "myorg" in your keychain.`
![image](https://github.com/jaraco/keyring/assets/447049/808485fe-37b1-4503-9471-d11e10fb8a02)
After entering the keyring password, the second dialog appears:
`python3.11 wants to access key "myorg" in your keychain.`
![image](https://github.com/jaraco/keyring/assets/447049/78a7cb3d-fb14-4def-b443-f96e0b47cc6f)

**To Reproduce**
1. Start Keychain Access application.
2. Create two passwords, where the name of one is a substring of the other name, e.g. `"myorg"`and `"myorg-api"`.
<img width="874" alt="image" src="https://github.com/jaraco/keyring/assets/447049/8fa4f69e-8b3c-4517-b117-2b73543b521a">

4. retrieve one of the passwords, e.g. `python -c 'import keyring;print(keyring.get_password("myorg", "user1"))'`
5. See two dialogs

**Expected behavior**
The keyring password should only be requested once.

**Environment**

- OS: macOS 13.4.1 (Ventura)
- Python: 3.11.4

```
keyring 24.2.0
...
keyring.backends.fail.Keyring (priority: 0)
keyring.backends.chainer.ChainerBackend (priority: -1)
keyring.backends.macOS.Keyring (priority: 5)
...
```
Strangely, when creating those entries in keyring using **python** and after deleting "python3.11" from "Access Control" within the [Keychain Access](https://github.com/jaraco/keyring#security-considerations) app, the keychain password is requested only once, as expected!

jaraco:
I suspect the issue is unrelated to the presence of a substring. I use the macOS keyring and it frequently prompts me twice for the keychain password with two different prompts.

To test this assumption, I added just one password to my keychain:

![image](https://github.com/jaraco/keyring/assets/308610/af04eb74-7b59-4162-aca0-169792f386d8)

[...]

Issue #619, opened 2023-01-18
Multiple Password Popups on macOS for Multiple Credentials When the Python Binary Updates


ChristopherHammond13:
**Describe the bug**
Firstly, I believe this may be related to #519?

When the Python binary updates (or a new virtual environment is created), I will receive a popup asking for my login/keychain password as usual. However, in my application (https://github.com/crowdstrike/Falcon-Toolkit) it is possible to configure many 'profiles', each of which will hold a separate secret (in the case of that application, it's an API Client Secret). Strangely, I will get the popup repeatedly, _one per profile_, so eight configurations with eight separate Client ID / Client Secret pairs will result in me needing to type my password _eight times_ even if I click "Allow All". Given that I work at a security company, and my password is therefore pretty complex, typing my password _n_ times every time I update Python or the virtual environment's Python binary is far from convenient :) I expect this to affect our MSSP and internal users a fair bit, too, if they use this tool across multiple client environments with separate sets of credentials.

I do not known if this is a bug in the way I am using keyring (I am always open to it being my fault!), keyring itself, or the macOS implementation of the binary validity checks. An example of the code I have that retrieves this value is here, in case this helps: https://github.com/CrowdStrike/Falcon-Toolkit/blob/main/falcon_toolkit/common/auth_backends/public_single_cid.py#L44. The code will run once per profile. I am using the Client ID as the username and the Client Secret as the password, with the service name defined as a constant in the code.

**To Reproduce**
1. Create an application that stores multiple named secrets to the macOS secrets store
2. Upgrade Python so that the binary changes
3. Attempt to retrieve / load all those secrets in one go
4. Observe one popup requesting the password for each loaded secret

I expect I can re-architect some of this code to defer the password request to later in the execution, but I still feel like there is a bug here as loading multiple credentials early in the execution is (in my opinion) quite a reasonable thing to do.

**Expected behavior**
I should be asked for my password once, and the "Allow All" button should stop future popups until I change the Python binary again.

**Environment**

- OS: macOS Ventura
- Virtual environments provided by pipx or Poetry (both are affected)

```
$ pip list | grep keyring
...
```
keyring 23.13.1

```
$ keyring --list-backends
...
```
```
keyring.backends.fail.Keyring (priority: 0)
keyring.backends.macOS.Keyring (priority: 5)
keyring.backends.chainer.ChainerBackend (priority: -1)
```

**Additional context**
[...]

Issue #512, opened 2021-05-11
Changing "Access Control" on macOS


arossert:
I'm building an app using PyInstaller and want to use keyring for secret management on Windows and macOS.
Everything is working as it should until I want to do an upgrade to the app, them the keychain on macOS ask for permissions to access the key.

I assume it is related to this #406 ticket, when the secret first added to the keychain it adds the binary file to the allowed applications under the "Access Control" tab and after an upgrade, it is a new binary.

Is there an option to set the "Allow all applications to access this item" settings the same as `security -A` works? 
I tried looking at the code but could not find how it can be done.

jaraco:
Short answer is, I don't know. Looking briefly at the [API docs](https://developer.apple.com/documentation/security/1398366-seckeychainaddgenericpassword/), there's no parameter to allow all. So probably the `security` tool is adding the credential and allowing the default behavior to take effect at first, then somehow allowing all binaries to have access. I wasn't able to figure out how to do that. I'd be delighted if someone with more macOS experience or time to invest would dive into the APIs and figure out what APIs need to be invoked and how that could be called from Python.

phatmandrake:
I'm using PyInstaller as well and are running into the same problem:

>  Keychain Services does not distinguish older and newer versions of your program as long as both are signed and the unique Identifier remains constant.

https://developer.apple.com/library/archive/documentation/Security/Conceptual/CodeSigningGuide/Procedures/Procedures.html

Indicates that as long as it's signed the same and the identifier doesn't change it should be treated the same by Keychain Services. I am still having issues, although I am using a self-signed certificate.

https://stackoverflow.com/questions/52705771/will-the-mac-prompt-the-user-again-to-allow-access-to-the-keychain-after-updatin

This is probably not a keyring specific issue though.

phatmandrake:
I finally got ahold of an Apple Developer ID Cert and Installer Cert and my pyinstaller application now behaves like the documentation suggest, whereas before with a self-signed cert it would prompt to re-authorize the app if it was replaced.

With that in mind "allowing all applications to access an item" sounds like a form of access control that keyring doesn't provide elsewhere (that I'm aware of), so it seems like that would be out of scope for the utility in its current form.

PR #746, opened 2026-03-11
macOS: use SecItemUpdate to preserve keychain ACL on password change

mdub:
## Summary

`set_generic_password` previously deleted and re-added keychain items on every update. This resets the item's access control list (ACL), wiping any "Always Allow" grants and causing repeated Keychain Access password prompts.

This PR uses `SecItemUpdate` to modify items in place, preserving the ACL. Falls back to delete+add when the item doesn't exist yet or the update is not possible (e.g. `SecItemUpdate` returns `errSecParam` for empty password data).

## Related issues

- #619 — Multiple Password Popups on macOS for Multiple Credentials When the Python Binary Updates
- #512 — Changing "Access Control" on macOS
- #644 — Two keyring password dialogs are triggered

## Changes

- Add `SecItemUpdate` binding alongside existing `SecItemAdd`/`SecItemDelete`
- Add `error.sec_param` (-50) and `InvalidParametersError` exception
- Extract `_update_generic_password` and `_add_generic_password` helpers
- `set_generic_password` tries update-in-place first, falls back to delete+add on `NotFound` or `InvalidParametersError`
- Add `FakeKeychain`-based tests verifying the update-first behaviour

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/keyring/backends/macOS/api.py
+++ b/keyring/backends/macOS/api.py
@@ -18,6 +18,7 @@ class error:
     item_not_found = -25300
     keychain_denied = -128
     sec_auth_failed = -25293
+    sec_param = -50
     plist_missing = -67030
     sec_interaction_not_allowed = -25308
 
@@ -56,6 +57,10 @@ SecItemCopyMatching.argtypes = (c_void_p, c_void_p)
 SecItemDelete = _sec.SecItemDelete
 SecItemDelete.restype = OS_status
 SecItemDelete.argtypes = (c_void_p,)
+
+SecItemUpdate = _sec.SecItemUpdate
+SecItemUpdate.restype = OS_status
+SecItemUpdate.argtypes = (c_void_p, c_void_p)
 
 CFDataGetBytePtr = _found.CFDataGetBytePtr
 CFDataGetBytePtr.restype = c_void_p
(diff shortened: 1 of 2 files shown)


mdub:
Closing this — testing showed that `SecItemUpdate` doesn't actually solve the popup problem I was trying to fix.

## What I found

The popups occur when **multiple Python binaries** (e.g. brew's Python 3.13 and a hermit/uv-managed Python 3.12) access the same keychain items for Snowflake SSO credential caching.

I tested whether `SecItemUpdate` preserves cross-binary "Always Allow" grants:

1. Binary A creates a keychain item → no popup
2. Binary B reads → popup → user clicks "Always Allow"
3. Binary A updates the value via `SecItemUpdate` → no popup
4. Binary B reads → **popup again**

macOS re-prompts other applications whenever the item's value changes, regardless of whether the update was done via `SecItemUpdate` or delete+add. This makes sense from a security perspective.

So while `SecItemUpdate` is arguably more idiomatic than delete+add, it doesn't help with the real-world problem that motivated this PR. The only practical solution for the multi-binary case is to avoid the macOS Keychain entirely (e.g. `PYTHON_KEYRING_BACKEND=keyrings.alt.file.PlaintextKeyring`).

Full investigation notes: https://github.com/squareup/personal-mikewilliams/tree/main/notes/2026-03-11-snowflake-cli-keyring-popup

PR #749, opened 2026-04-09
Pin GitHub Actions to full-length commit SHAs

r7sy:
Pin all GitHub Actions in CI workflows to their full-length commit SHAs using [pinact](https://github.com/suzuki-shunsuke/pinact).

### Changes

- `actions/checkout@v4` → `actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1`
- `actions/setup-python@v5` → `actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0`

### Rationale

GitHub Action tags are mutable — a tag like `v4` can be force-pushed to point to a different commit at any time. Pinning to a full-length commit SHA ensures workflows use an immutable reference, mitigating the risk of a compromised or hijacked upstream action.

This is a [GitHub-recommended security practice](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions):

> Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release.

### Benefit

This change improves the project's [OpenSSF Scorecard](https://securityscorecards.dev/) score on the **Pinned-Dependencies** check, which verifies that workflow dependencies are pinned to specific, immutable versions. The version tag is preserved as a comment (e.g., `# v4.3.1`) for readability and to support automated update tools like Renovate and Dependabot.


r7sy:
@mitya57 sorry for ping but would appreciate it if this could be reviewed, hopefully it is low hanging fruit. It unblocks my org from using the module, so it is really helpful for us.

mitya57 (approved):
LGTM, but the final word will be from @jaraco.

r7sy:
> LGTM, but the final word will be from @jaraco.

@mitya57 thank you for the initial approval.

@jaraco is it possible to push this one through? The scope is fairly limited.

PR #751, opened 2026-04-21
Add ssh-agent-keyring to third-party backends

monperrus:
## Summary
- add ssh-agent-keyring to the Third-Party Backends list in README.rst
- keep the entry wording consistent with the existing backend descriptions

This updates the project README that is rendered on PyPI.


andraxin:
Please consider https://github.com/monperrus/ssh_agent_keyring/issues/2 before accepting this PR!

monperrus:
Thanks for looking into this. Addressed both problems.

andraxin:
> Thanks for looking into this. Addressed both problems.

Actually, https://github.com/monperrus/ssh_agent_keyring/issues/2 raised 3 problems, not 2.

1. Broken -- FIXED (although with an AI-level brittle approach)
2. Unsafe -- FIXED? (sort of, better options are available)
3. Ultimately Unfixable -- NOT FIXED (explained away, in documentation)

At the very least, I'd expect an update to the suggested changes here
that point out that the security guarantees afforded by this backend
differ substantially from most other backends.

monperrus:
Third concern solved in https://github.com/monperrus/ssh_agent_keyring/pull/5

andraxin:
> Third concern solved in [monperrus/ssh_agent_keyring#5](https://github.com/monperrus/ssh_agent_keyring/pull/5)

I can only hope that this dishonest response was AI-generated.
The 3rd concern was NOT solved, since it cannot ever be solved.
On an untrusted system, intercepting communication with the agent
will ALWAYS be possible.

Granted, the cited fix _does_ raise the bar and eliminates simple
`PATH`-based wrappers, which makes the proposed backend's
security comparable to https://github.com/eszio/age-plugin-sshagent

Unfortunately (and, predictably), the agent socket can still be interposed
as the example in  https://github.com/monperrus/ssh_agent_keyring/issues/2#issuecomment-5382981552 shows.

PR #758, opened 2026-05-31
Add keyrings.envvars backend to README.rst

### LINKED ISSUE #539: New backend for getting credentials from environment variables
I'm planning to implement a keyring backend for getting credentials from environment variables.  This will be useful in CI environments (eg. Jenkins) for easily passing credentials from the CI credential store to pip and other tools that support `keyring`.

The current plan is to use environment variable format like this:
```bash
KEYRING_SERVICE_NAME_0=https://index0.example.com
KEYRING_SERVICE_USERNAME_0=myusername
KEYRING_SERVICE_PASSWORD_0=mypassword
KEYRING_SERVICE_NAME_1=https://index1.example.com
KEYRING_SERVICE_USERNAME_1=myusername1
KEYRING_SERVICE_PASSWORD_1=mypassword1
```
Is this format acceptable, or should I avoid using the `KEYRING_` prefix?

I don't think there is any need to allow keyring to set any username or password environment variables through this backend, so it would end up being a read-only backend supporting the `get` operation but not `set` or `del`.

Are there any potential issues I've missed with an implementation like this?

wwuck:
Add keyrings.envvars backend into the third-party backend list in README.rst 

Closes #539

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -193,6 +193,7 @@ use cases. Simply install them to make them available:
 - `keyring_jeepney <https://pypi.org/project/keyring_jeepney>`__ - a
   pure Python backend using the secret service DBus API for desktop
   Linux (requires ``keyring<24``).
+- `keyrings.envvars <https://pypi.org/project/keyrings.envvars>`__ - a backend that retrieves credentials from environment variables.
 
 
 Write your own keyring backend


wwuck:
@jaraco are there any other changes needed with the failed checks, or are they unrelated to this?

PR #764, opened 2026-08-01
Fix: prevent KWallet from triggering creation UI on read queries (#496)

### LINKED ISSUE #496: Backend inference between KWallet and Gnome leads to unexpected results
> The SecretService keyring will be queried first, but if it's a miss (no password defined), the kwallet backend(s) will be queried, triggering the wallet creation UI. I believe the SecretService store is being queried first, but just not satisfying the operation. [...]
> In some ways, the behavior you describe is by design. But you've stumbled on a flaw in the design. [...]
> [...] Where a backend triggers blocking UI elements and [raises KeyringLocked exceptions](https://github.com/jaraco/keyring/blob/a65aa2ea949e78f67dc03abc4bfd4a691bdf0802/keyring/backends/kwallet.py#L101), however, that's not desirable.

My scenario is the same described by the OP, including _exactly_ the same set of backends and priorities: Gnome user, kwallet installed but never used (possibly pulled as a dependency of some other software), on first run (i.e, when trying to read data that was not yet set) it triggers a KDE Wallet Creating UI.

I realize that after some data is set (possibly to Gnome Keyring, given my backend priorities) the KDE UI will not show anymore. But still, this first-time UI is _very_ annoying and surprising.

What's the current status on this? I mean, what should I do, as a third-party developer, to prevent this from happening with my end-users? Is there any "official" (or "blessed") workaround?

A possible approach I'm thinking for my specific software is to try to detect a first-time usage by other means (a missing config file, for example) and the set the data to a dummy/blank value before attempting to read it. That, I _think_, would avoid triggering the "backend found but no data set, let's try another backend" behavior.

Is this a good approach? Any recommendations?

_Originally posted by @MestreLion in https://github.com/jaraco/keyring/issues/391#issuecomment-773878621_

newchannelid432-code:
Fixes #496.

## Problem
As discussed in #496, when a user is on a Gnome desktop environment but has \`kwallet\` installed (often pulled as a dependency by unrelated software), the \`ChainerBackend\` queries \`SecretService\` first. Since the password hasn't been set yet, it registers as a 'miss'. The \`Chainer\` then falls back to \`kwallet\`.

Because KWallet has never been initialized in this Gnome environment, querying it via DBus unconditionally calls \`self.iface.open()\`. This **triggers the KDE Wallet creation UI**, which is extremely annoying and surprising to the user who never intended to use KDE.

## Solution
This PR implements the exact solution requested by @jaraco in #496:
1. Adds a \`create=True\` flag to \`connected()\`.
2. \`get_password()\` and \`get_credential()\` now pass \`create=False\`.
3. If \`create=False\`, \`connected()\` queries \`self.iface.wallets()\` to check if the target wallet exists. If it does not, it returns \`False\` immediately **without** calling \`open()\`, completely preventing the UI from triggering.
4. \`get_password\` and \`get_credential\` now return \`None\` instead of raising \`KeyringLocked\` when the wallet doesn't exist. This allows the \`Chainer\` to gracefully move to the next backend without crashing the application.

This ensures that KWallet only presents its initialization UI when a user is explicitly *setting* a password, perfectly resolving the first-time usage flaw on mixed desktop environments.


keyring/backends/kwallet.py
```diff
@@ -94,7 +94,17 @@ def connected(self, service):
         try:
```
kramlipi, line 94: **Should fix** · design · `keyring/backends/kwallet.py:94` · Fragile DBus exception handling

What: The `connected` method catches `dbus.DBusException` and silently passes, falling back to `open()`. Why it matters: This masks potential connection or configuration issues, making debugging difficult if the wallet is missing or inaccessible for reasons other than the one intended. Do this: Log the exception or at least differentiate between 'wallet not found' and 'DBus communication failure'.

**Callers that may care:**
- `NEWS.rst:645` → `connected` — `directly from stdin if stdin is connected to a pipe.`
- `keyring/backend.py:40` → `__init__` — `super().__init__(name, bases, dict)`
- `keyring/errors.py:49` → `__init__` — `self.exc_info.__init__(*exc_info)`
- `NEWS.rst:349` → `__init__` — `to invoke ``super().__init__()``.`
- `tests/backends/test_kwallet.py:42` → `get_password` — `assert keyring.get_password(service, username) is None`

keyring/backends/kwallet.py
```diff
@@ -94,7 +94,17 @@ def connected(self, service):
...
+            wallet_name = self.iface.networkWallet()
+            if not create:
+                try:
+                    if wallet_name not in self.iface.wallets():
+                        return False
+                except dbus.DBusException:
+                    pass  # Fallback to open() if wallets() fails
+
+            self.handle = self.iface.open(wallet_name, wId, self.appid)
         except dbus.DBusException as e:
```
kramlipi, line 108: **Should fix** · logic · `keyring/backends/kwallet.py:108` · Silent failure on locked keyring

What: `get_password` returns `None` if `self.handle < 0` after a failed connection attempt. Why it matters: This behavior is inconsistent with the subsequent `KeyringLocked` exception. If the keyring is locked, the caller might expect an error rather than a silent `None`, which could be misinterpreted as 'password not found'. Do this: Clarify if `None` is the intended return value for a locked state or if it should propagate an error.

**Callers that may care:**
- `NEWS.rst:645` → `connected` — `directly from stdin if stdin is connected to a pipe.`
- `keyring/backend.py:40` → `__init__` — `super().__init__(name, bases, dict)`
- `keyring/errors.py:49` → `__init__` — `self.exc_info.__init__(*exc_info)`
- `NEWS.rst:349` → `__init__` — `to invoke ``super().__init__()``.`
- `tests/backends/test_kwallet.py:42` → `get_password` — `assert keyring.get_password(service, username) is None`

keyring/backends/kwallet.py
```diff
@@ -84,7 +84,7 @@ def _migrate(self, service):
             if not entry_list:
                 self.iface.removeFolder(self.handle, old_folder, self.appid)
 
-    def connected(self, service):
+    def connected(self, service, create=True):
```
kramlipi, line 87: **Worth a look** · api_break · `keyring/backends/kwallet.py:87` · No behavior risk: api_signature

**What:** No behavior risk: api_signature

**Why it matters:** The addition of the `create` parameter to `connected` is internal to the `DBusKeyring` class and does not affect the public API of the `keyring` package.

**Do this:** Update callers in the blast radius, or keep a compat shim and note it in the PR.

**Evidence:** `def connected(self, service, create=True):`

**Callers that may care:**
- `NEWS.rst:645` → `connected` — `directly from stdin if stdin is connected to a pipe.`
- `keyring/backend.py:40` → `__init__` — `super().__init__(name, bases, dict)`
- `keyring/errors.py:49` → `__init__` — `self.exc_info.__init__(*exc_info)`
- `NEWS.rst:349` → `__init__` — `to invoke ``super().__init__()``.`
- `tests/backends/test_kwallet.py:42` → `get_password` — `assert keyring.get_password(service, username) is None`

keyring/backends/kwallet.py
```diff
@@ -105,7 +115,9 @@ def connected(self, service):
 
     def get_password(self, service, username):
```
kramlipi, line 116: **Worth a look** · vulnerability · `keyring/backends/kwallet.py:116` · No behavior risk: auth

**What:** No behavior risk: auth

**Why it matters:** The change prevents the KWallet UI from popping up unexpectedly during read operations, which improves the user experience without weakening the underlying authentication mechanism.

**Do this:** Replace with a safe API (parameterized / no shell / no eval) and add a regression test.

**Evidence:** `def get_password(self, service, username):`

**Callers that may care:**
- `NEWS.rst:645` → `connected` — `directly from stdin if stdin is connected to a pipe.`
- `keyring/backend.py:40` → `__init__` — `super().__init__(name, bases, dict)`
- `keyring/errors.py:49` → `__init__` — `self.exc_info.__init__(*exc_info)`
- `NEWS.rst:349` → `__init__` — `to invoke ``super().__init__()``.`
- `tests/backends/test_kwallet.py:42` → `get_password` — `assert keyring.get_password(service, username) is None`

kramlipi, line 116: **Worth a look** · design · `keyring/backends/kwallet.py:116` · No blast-radius risk

**What:** No blast-radius risk

**Why it matters:** The changes are localized to the `kwallet` backend and the internal `connected` method, which is not part of the public API surface used by external callers.

**Do this:** Either fix it in this PR, or reply here explaining why the current behavior is intentional.

**Evidence:** `keyring/core.py:65`

**Callers that may care:**
- `NEWS.rst:645` → `connected` — `directly from stdin if stdin is connected to a pipe.`
- `keyring/backend.py:40` → `__init__` — `super().__init__(name, bases, dict)`
- `keyring/errors.py:49` → `__init__` — `self.exc_info.__init__(*exc_info)`
- `NEWS.rst:349` → `__init__` — `to invoke ``super().__init__()``.`
- `tests/backends/test_kwallet.py:42` → `get_password` — `assert keyring.get_password(service, username) is None`

kramlipi (commented):
Automated code review by **kramlipi code-agent**.
---
*Review notes from **kramlipi code-agent** (@kramlipi_code_review_bot) — grounded scanners + critic, not a rubber stamp.*

**Free for open-source / public repos.** Ready-made GitHub Actions: [code-review CI setup](https://kramlipi.github.io/code-review-ci/) · Product: [www.kramlipi.com](https://www.kramlipi.com) · Docs: [https://kramlipi.github.io/](https://kramlipi.github.io/)

kramlipi:
### Review notes

**5 note(s)** — 0 must-fix · 2 should-fix · 3 worth a look

Themes: design (2), logic (1), api_break (1), vulnerability (1)

Start here:
- `keyring/backends/kwallet.py:94` (should-fix) — Fragile DBus exception handling
- `keyring/backends/kwallet.py:108` (should-fix) — Silent failure on locked keyring
- `keyring/backends/kwallet.py:87` (look) — No behavior risk: api_signature
- `keyring/backends/kwallet.py:116` (look) — No behavior risk: auth
- `keyring/backends/kwallet.py:116` (look) — No blast-radius risk
---
*Review notes from **kramlipi code-agent** (@kramlipi_code_review_bot) — grounded scanners + critic, not a rubber stamp.*

**Free for open-source / public repos.** Ready-made GitHub Actions: [code-review CI setup](https://kramlipi.github.io/code-review-ci/) · Product: [www.kramlipi.com](https://www.kramlipi.com) · Docs: [https://kramlipi.github.io/](https://kramlipi.github.io/)

PR #767, opened 2026-08-15
Fix empty piped password falling through to getpass

VXNCXNX:
Setting an empty password from a pipe crashes instead of storing it.

```
$ printf '' | keyring set svc usr
  File ".../getpass.py", line 148, in _raw_input
    raise EOFError
EOFError
```

`printf '\n' | keyring set svc usr` fails the same way, and on a terminal without a controlling tty it surfaces as `termios.error` first.

## Cause

```python
return self.pass_from_pipe() or getpass.getpass(prompt)
```

`pass_from_pipe` returns `False` when there is no pipe, and the piped string otherwise. An empty pipe therefore returns `''`, which is falsy but is not `False`, so the `or` falls through to interactive `getpass` on a stdin that is not a tty. That is the one situation where `getpass` cannot succeed.

The docstring already states the two-valued contract:

```python
"""Return password from pipe if not on TTY, else False."""
```

so the fix is to test for `False` rather than for falsiness.

## Verification

End to end with a real backend:

```
$ printf '' | keyring --keyring-backend keyrings.alt.file.PlaintextKeyring set svc usr
$ keyring --keyring-backend keyrings.alt.file.PlaintextKeyring get svc usr

$ echo $?
0
```

The empty password is stored and read back as empty, rather than the command failing.

The other two paths are unchanged: `printf 'secret\n' | keyring set ...` still stores `secret`, and with a tty it still prompts.

`test_set_pipe_empty` in `tests/test_cli.py`, parametrised over `''` and `'\n'`, patches `getpass` to emit a distinct value so an accidental fallback is visible in the assertion rather than hanging.

Reverting only the condition, back to `piped or getpass.getpass(prompt)`, fails it as an assertion rather than an import error:

```
E   AssertionError: expected call not found.
E   Expected: set_password('svc', 'usr', '')
E     Actual: set_password('svc', 'usr', 'prompted')
```

`pytest tests/` is 25 passed, 94 skipped, 2 xpassed. ruff-format is clean.

Two pre-existing items left alone, both confirmed identical with my change stashed: a `BLE001` ruff finding, and a mypy error about the `shtab` import.

*Disclosure: written with AI assistance (Claude Code). I reproduced the crash, verified the round trip against a real backend, checked all three input paths, and ran the revert check myself.*


VXNCXNX:
Open a month with a red X that no longer has anything behind it, so here is the current state.

The branch is level with `main` and the suite is green locally:

```
$ pytest
36 passed, 94 skipped, 2 xpassed
```

The failing `tests` run was on 15 August and its jobs have since been purged, so the log is gone and there is nothing left to read. I could not reproduce a failure on the current base.

The change is `+16/-1`. Setting an empty password from a pipe crashes rather than storing it:

```
$ printf '' | keyring set svc usr
EOFError
```

`pass_from_pipe()` returns `False` when there is no pipe and the piped string otherwise, so an empty pipe returns `''`. That is falsy but not `False`, so `or` falls through to interactive `getpass` on a stdin that is not a tty, which is the one case where `getpass` cannot succeed.

